PR #80, opened 2022-06-15
6 archive@main

### LINKED ISSUE #6: Implement the `Storager` class
Implement a class responsible for saving a `Report` to a format that can be used to restore the `Report` object to a previous state.

The motivation is: 
* we would like to have an option of saving and restoring `Reporter` sessions, also make it easier to share reports 

Some suggestions:
* The report format should be self-contained - if any other files are needed, they should be included (it might be a .zip file!)
* store key-value data in a json or .rds

Here is how it fits into the larger picture: https://app.diagrams.net/#G1MOUaQWakvtz0D_S6VlhHY8j8qzDu1qCW

Polkas:
closes #6 

the new issue to create a module and implement it in vignettes (https://github.com/insightsengineering/teal.reporter/issues/81)

**The functional objective is to be able to save the report state and to retrieve it from saved zip file.**

No influence on the old code, the old tests are not touched and are still passing.

Scope:

- `Archiver` prototype class with 2 main methods `read` and `write`.
- `JSONArchiver` implementation of the `FileArchiver` class which is an implementation of `Archiver`, we want to be able to convert reporter to zip file with JSON and static files or convert this zip file to reporter back.
- metadata in `Reporter` with setter and getter
- `from_reporter` method in the `Reporter` class
- `to_jsondir` and `from_jsondir` methods in the `Reporter` class
- new class `FileBlock` which is a parent for the `Picture` and `Table` Blocks. Helps to reduce the code lines. 
- Two additional methods `from_list` and `to_list` in the Blocks and ReportCard.
- fix a few docs problems

We support only the Archiver versions we want to, in the `from_jsondir` Reporter method. Only one version of the Archiver is available for `write` and a certain package version. If anything influential is updated in the `to_jsondir` method then the version should be updated too.

Example:

```
card1 <- teal.reporter:::ReportCard$new()

card1$append_text("Header 2 text", "header2")
card1$append_text("A paragraph of default text", "header2")
card1$append_plot(
 ggplot2::ggplot(iris, ggplot2::aes(x = Petal.Length)) + ggplot2::geom_histogram()
)

card2 <- teal.reporter:::ReportCard$new()

card2$append_text("Header 2 text", "header2")
card2$append_text("A paragraph of default text", "header2")
lyt <- rtables::analyze(rtables::split_rows_by(rtables::basic_table(), "Day"), "Ozone", afun = mean)
table_res2 <- rtables::build_table(lyt, airquality)
card2$append_table(table_res2)
card2$append_table(iris)

reporter <- teal.reporter:::Reporter$new()
reporter$append_cards(list(card1, card2))
```

```
archiver <- JSONArchiver$new()
archiver$write(reporter)

archiver$read()$get_cards()[[1]]$get_content()
archiver$get_output_dir()

archiver$read(archiver$get_output_dir())$get_cards()[[1]]$get_content()[[3]]$get_content()
```

```
archiver <- JSONArchiver$new()
archiver$write(reporter)
Reporter$new()$from_reporter(archiver$read())
doc <- Renderer$new()$render(Reporter$new()$from_reporter(archiver$read())$get_blocks())
```


R/Archiver.R
```diff
@@ -0,0 +1,163 @@
...
+    }
+  ),
+  lock_objects = TRUE,
+  lock_class = TRUE
+)
+
+#' @title `RDSArchiver`
+#' @keywords internal
+FileArchiver <- R6::R6Class( # nolint: object_name_linter.
+  classname = "RDSArchiver",
```
kpagacz, line 34: Classname does not match the binding name.

Polkas: thanks

R/Reporter.R
```diff
@@ -159,16 +161,165 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
+              TextBlock = TextBlock$new()$from_list(block),
+              PictureBlock = {
+                new_file_path <- tempfile(fileext = ".png")
+                file.copy(file.path(output_dir, basename(block$path)), new_file_path)
+                PictureBlock$new()$from_list(list(path = new_file_path))
+              },
+              TableBlock = {
+                new_file_path <- tempfile(fileext = ".RDS")
+                file.copy(file.path(output_dir, basename(block$path)), new_file_path)
+                TableBlock$new()$from_list(list(path = new_file_path))
```
Polkas, line 299: Very important part. 
When the FileBlock is finalized then its content path is removed so we have to be careful here.
We have to use an external files for the PictureBlock and TableBlock content paths. If we add here the archive paths then we will break the process, as then the archive files will be removed when the blocks will be removed.

R/FileBlock.R
```diff
@@ -0,0 +1,50 @@
...
+    #' block$from_list(list(path = file_path))
+    from_list = function(x, output_dir) {
+      checkmate::assert_list(x)
+      checkmate::assert_names(names(x), must.include = "basename")
+      path <- file.path(output_dir, x$basename)
+      file_type <- paste0(".", tools::file_ext(path))
+      checkmate::assert_file_exists(path, extension = file_type)
+      new_file_path <- tempfile(fileext = file_type)
+      file.copy(path, new_file_path)
+      super$set_content(new_file_path)
```
Polkas, line 33: Very important part.
When the FileBlock is finalized then its content path is removed so we have to be careful here.
We have to use an external files for the PictureBlock and TableBlock content paths. If we add here the archive paths then we will break the process, as then the archive files will be removed when the blocks will be removed.

R/TableBlock.R
```diff
@@ -27,18 +27,12 @@ TableBlock <- R6::R6Class( # nolint: object_name_linter.
     #' block$set_content(iris)
     #'
     set_content = function(content) {
+      is.character(content)
```
nikolas-burkoff, line 30: something missing here?

R/TextBlock.R
```diff
@@ -53,6 +53,28 @@ TextBlock <- R6::R6Class( # nolint: object_name_linter.
...
+      self$set_style(x$style)
+      invisible(self)
+    },
+    #' @description Convert the `TextBlock` to a list.
+    #' @return `named list` with a text and style.
+    #' @examples
+    #' block <- teal.reporter:::TextBlock$new()
+    #' block$to_list()
+    to_list = function() {
+      list(text = super$get_content(), style = self$get_style())
```
nikolas-burkoff, line 77: As discussed I think this should be `self$get_content()`

tests/testthat/test-Renderer.R
```diff
@@ -1,33 +1,33 @@
 testthat::test_that("Renderer object can be created", {
-  testthat::expect_error(Renderer$new(), regexp = NA)
+  testthat::expect_error(teal.reporter::Renderer$new(), regexp = NA)
```
nikolas-burkoff, line 2: you shouldn't need to use teal.reporter:: in the tests ?? See e.g. https://github.com/insightsengineering/teal.slice/blob/b0c6dafb0410371fa54b7db13b962fdbabccc6d3/tests/testthat/test-FilteredData.R#L6

R/TableBlock.R
Polkas: rm

tests/testthat/test-Renderer.R
Polkas: I think it is a good practice

R/TextBlock.R
Polkas: done

R/TextBlock.R
```diff
@@ -53,6 +53,28 @@ TextBlock <- R6::R6Class( # nolint: object_name_linter.
...
+    #' @param x `named list` with two fields `c("text", "style")`.
+    #' Use the `get_available_styles` method to get all possible styles.
+    #' @return invisibly self
+    #' @examples
+    #' block <- teal.reporter:::TextBlock$new()
+    #' block$from_list(list(text = "sth", style = "default"))
+    from_list = function(x) {
+      checkmate::assert_list(x)
+      checkmate::assert_names(names(x), must.include = c("text", "style"))
+      super$set_content(x$text)
```
nikolas-burkoff, line 67: also here :)

Polkas: done

tests/testthat/test-Renderer.R
nikolas-burkoff: I'll leave this one to you

kpagacz (commented):
I would love to see from_list and to_list on the Reporter level as well. This way it is clear what is the medium of communication between `Archiver` and `Reporter`. 

This is important because we might want to attach some `Reporter`-level metadata that should not be handled by `Archiver` because `Reporter` owns the metadata and not the `Archiver`.

The version checking should happen inside `Reporter` because otherwise in the future there will be no place to mitigate the versions in a sensible way. You might even think about the read method only reading in the source and returning the list and then `Reporter` can consume this list via `from_list`. This creates space for other tools to work on this list in case we need it in the future. It also lets users peek at this list at their leisure and decouple `Archiver` from `Reporter`.

I am unsure about the new `Block` class. I am convinced the new method could live inside `ContentBlock` considering its copy constructor is already aware that content can be a path to a file and this would basically remove any need for it.

I am also struggling the understand the need to split `Archiver` and `FileArchiver`. I am aware that we want to declare the interface but it seems that either we should move the logic from `FileArchiver` to `JSONArchiver` or to `Archiver` and do not create long, pointless inheritance trees.

Polkas:
@kpagacz some of your requirements was already there so it was very easy for me to retrieve them. 
Reporter has own to_list/from_list now.

I am sure i want to have FILEArchiver as we think about YAMLArchiver and others so it will be easy to implement them now.
I feel the same about FileBlock that we could add other file connected objects other than Table and Plot.

R/Archiver.R
```diff
@@ -0,0 +1,166 @@
+#' @title `Archiver`
+#' @keywords internal
+Archiver <- R6::R6Class( # nolint: object_name_linter.
+  classname = "Archiver",
+  public = list(
+    #' @description Returns a `Archiver` object.
+    #'
+    #' @return a `Archiver` object
```
mhallal1, line 8:
```suggestion
    #' @description Returns an `Archiver` object.
    #'
    #' @return an `Archiver` object
```

kpagacz:
Given the implementation of Reporter$to_list, we can decouple Archiver and Reporter completely. Here is a diagram of how this could go down: https://drive.google.com/file/d/1EXmD72mqg4jVeYfyrDdybgoEGtr7O3Lr/view?usp=sharing

Polkas:
I just transfer the to_jsondir/from_jsondir from Archiver to Reporter, some how revert https://github.com/insightsengineering/teal.reporter/commit/60ff11e288777d020c4e71fe932c7d2b5d566b38.
Additional tests were added.

Polkas:
We had a talk to update the `Archiver$write` method to consume the list instead of reporter.
I tried to do it. Unfortunately it break a consistency as we have to manually operate with directories then.
The proposed update force the manual creation of the directory provided to Reporter$to_list() which next have to be provided to the Archiver. The process is now automatic.
Thus user have to manipulate with the directories manually and the manually created directories are not automatically destroyed when the Archiver is removed. As there is no owner of the directory (currently output_dir in the Archiver) nobody is responsible for it.

kpagacz (approved):
Just be aware that this implementation will most likely need to change when there comes a time of attempts to "migrate" the json representation. Otherwise, it will be grossly inefficient (many unneded reads and writes to the file system).

PR #97, opened 2022-07-04
append_fs accepts a list

### LINKED ISSUE #96: TealReportCard - append_fs should consume a simple list
Shortly we will stop to use datasets inside the modules, so we have to update the append_fs in the TealReportCard.
It should consume a simple list and internally some styling function like `yaml::as.yaml` have to be used to format it to the nice printout.

Polkas:
closes #96 

`append_fs` in the `TealReportCard` accept a regular `list` now.

The drawback is that under yaml::as.yaml ranges are not so clear because of lack of the from/to words. Another thing is a longer printout.

Because of this too long printout we should think about sth custom.

### DIFF
--- a/R/TealReportCard.R
+++ b/R/TealReportCard.R
@@ -28,22 +28,18 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
     #' @description Appends the filter state list to the `content` and `metadata` of this `TealReportCard`.
     #'
     #' @param fs (`list`) a filter states.
-    #' @param format_fun (`function`) a function to format the filter states.
     #' @return invisibly self
     #' @examples
     #' card <- TealReportCard$new()$append_fs(list(a = 1, b = 2))
     #' card$get_content()[[1]]$get_content()
     #'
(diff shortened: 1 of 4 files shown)


Polkas:
@mhallal1 I think we have to first choose the solution https://github.com/insightsengineering/teal.reporter/issues/96#issuecomment-1173724686

R/TealReportCard.R
```diff
@@ -25,39 +25,39 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
       super$append_metadata("SRC", src)
       invisible(self)
     },
-    #' @description Appends the filter state list to the `content` meta data  of this `TealReportCard`.
+    #' @description Appends the filter state list to the `content` and `metadata` of this `TealReportCard`.
     #'
-    #' @param fs (`FilteredData`) with filter states.
+    #' @param fs (`list`) a filter states.
```
mhallal1, line 30:
```suggestion
    #' @param fs (`list`) of filter states.
```

mhallal1 (approved):

PR #110, opened 2022-07-06
Handle report generation for CI

dinakar29:
The XML reporting ought to happen during CI runs.


dinakar29:
I think I'll propagate this across all repos. WDYT?

nikolas-burkoff:
> I think I'll propagate this across all repos. WDYT?
That's fine with me - @gogonzo ?

PR #112, opened 2022-07-11
add support for heatmap

mhallal1:
part of https://github.com/insightsengineering/teal.modules.hermes/issues/238
 Add support for `Heatmap` class in `PictureBlock` to  support pca module `tmh`.

### DIFF
--- a/R/PictureBlock.R
+++ b/R/PictureBlock.R
@@ -40,7 +40,7 @@ PictureBlock <- R6::R6Class( # nolint: object_name_linter.
           if (inherits(content, "grob")) {
             grid::grid.newpage()
             grid::grid.draw(content)
-          } else if (inherits(content, "gg") || inherits(content, "Heatmap")) {
+          } else if (inherits(content, c("gg", "Heatmap")) {
             print(content)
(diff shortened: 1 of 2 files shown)


Polkas:
Please add some tests for it, the needed package could be in Suggests.

Polkas (approved):
As `ComplexHeatmap::Heatmap` demands a Bioc dependency so I think we could skip it or artificially return an object of its type.

R/PictureBlock.R
```diff
@@ -40,7 +40,7 @@ PictureBlock <- R6::R6Class( # nolint: object_name_linter.
           if (inherits(content, "grob")) {
             grid::grid.newpage()
             grid::grid.draw(content)
-          } else if (inherits(content, "gg")) {
+          } else if (inherits(content, "gg") || inherits(content, "Heatmap")) {
```
gogonzo, line 43: unlike `is`, `inherits` accept a vector 
```suggestion
          } else if (inherits(content, c("gg", "Heatmap")) {
```

mhallal1:
replaced by https://github.com/insightsengineering/teal.reporter/pull/113

PR #119, opened 2022-07-20
update append_fs

mhallal1:
related to https://github.com/insightsengineering/teal.slice/issues/61

update `append_fs` so that list is used to append meta data in the reporter and the character output of the attr of `fs` is used to append text to the reporter.

Please test with: https://github.com/insightsengineering/teal.slice/pull/75

### DIFF
--- a/R/TealReportCard.R
+++ b/R/TealReportCard.R
@@ -35,11 +35,7 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
     #'
     append_fs = function(fs) {
       checkmate::assert_list(fs)
-      super$append_text(yaml::as.yaml(fs, handlers = list(
-        POSIXct = function(x) format(x, "%Y-%m-%d"),
-        POSIXlt = function(x) format(x, "%Y-%m-%d"),
-        Date = function(x) format(x, "%Y-%m-%d")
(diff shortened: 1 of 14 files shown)


R/TealReportCard.R
```diff
@@ -35,11 +35,7 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
     #'
     append_fs = function(fs) {
       checkmate::assert_list(fs)
-      super$append_text(yaml::as.yaml(fs, handlers = list(
-        POSIXct = function(x) format(x, "%Y-%m-%d"),
-        POSIXlt = function(x) format(x, "%Y-%m-%d"),
-        Date = function(x) format(x, "%Y-%m-%d")
-      )), "verbatim")
+      super$append_text(attr(fs, "formatted"), "verbatim")
       super$append_metadata("FS", fs)
```
nikolas-burkoff, line 39: Do you actually need super here? I would have thought self$ would have worked unless this classes append_text and append_metadata has overridden the base class implementation and we specifically want the base class implementation

mhallal1: updated.

Polkas (approved):
LGTM

PR #132, opened 2022-08-23
Update check.yaml

arkadiuszbeer:
# Pull Request

Fixes #nnn


pawelru:
@Polkas can you please enter this feature branch and fix the notes detected here? At least one requires a simple fix.

Polkas:
> @Polkas can you please enter this feature branch and fix the notes detected here? At least one requires a simple fix.

These two notes will be in each repo so we have to have a consistent update. BTW. It is only a hint which not have to be applied, not sure "We recommend to" sounds better. 

@nikolas-burkoff

gogonzo (approved):

PR #143, opened 2022-10-06
Make document types names more friendly

### LINKED ISSUE #142: Change dropdown labels for document type
I wonder whether it would be better to change dropdown menu to something less technical: html / pdf / powerpoint / word, instead of html_document / powerpoint_presentation etc.

image

nikolas-burkoff:
Closes #142 

Test with apps in the vignettes

image
image


BLAZEWIM (approved):
Checked, works fine.

PR #148, opened 2022-10-10
Update Renderer.R

### LINKED ISSUE #147: Reporter: table resizing issue

Polkas:
closes #147 

Not the height of the plot will be taken in to account too when it is rendered.

Please render any html document and check the img tag that now it contains both height and width. 

image


Polkas:
this fix the html but broken the pdf one.

Issue #137, opened 2022-10-04
Freeze Add Card button when adding card is in progress


nikolas-burkoff:
I assume for backlog? 

When adding a card takes a long time you get the following screen:

![image](https://user-images.githubusercontent.com/15201933/193833062-5a306a23-b5af-44b3-a538-03dd4f68f794.png)

But you can still click on the "Add Card button" which leads to multiple cards being added:

![image](https://user-images.githubusercontent.com/15201933/193833260-9a2a7fae-9abc-4b95-9ded-5af942fda76c.png)

donyunardi:
> I assume for backlog?
Yes, moving to backlog.

donyunardi:
As @nikolas-burkoff pointed out, this is taken care with #151.
Closing.

PR #151, opened 2022-10-12
disable the add card button when clicked

Polkas:
closes https://github.com/insightsengineering/coredev-tasks/issues/400

Fix the problem of slow computation and possible multi added cards for multi clicks.

Please add Sys.sleep(10) to card_fun during test to simulate slow evaluation, e.g. in the previewer example app.

### DIFF
--- a/R/AddCardModule.R
+++ b/R/AddCardModule.R
@@ -8,9 +8,27 @@
 #' @export
 add_card_button_ui <- function(id) {
   ns <- shiny::NS(id)
+
   shiny::tagList(
     shiny::singleton(
       shiny::tags$head(shiny::includeCSS(system.file("css/custom.css", package = "teal.reporter")))
+    ),
+    shiny::singleton(
+      shiny::tags$head(
+        shiny::tags$script(
+          shiny::HTML(
+            sprintf(
+              '
+              $(document).ready(function(event) {
+                $("body").on("click", "#%s", function() {
+                  $(this).addClass("disabled");
+                })
+              })',
+              ns("add_card_ok")
+            )
+          )
+        )
+      )
     ),
     shiny::tags$button(
       id = ns("add_report_card_button"),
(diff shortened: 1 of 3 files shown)


nikolas-burkoff:
Also closes https://github.com/insightsengineering/teal.reporter/issues/137 :)

R/AddCardModule.R
```diff
@@ -8,10 +8,28 @@
...
+                $("body").on("click", "#%s", function() {
+                  $(this).addClass("disabled");
+                })
+              })',
+              ns("add_card_ok")
+            )
+          )
+        )
+      )
+    ),
```
gogonzo, line 32: instead of js and css code you can also do following 
```r
      shiny::observeEvent(input$add_card_ok, {
        shinyjs::disable("add_card_ok")
```

I prefer this as it stays within R code - easier to review in the future

Polkas: shinyjs is not a dependency on purpose

Polkas: We could add a comment here to help

gogonzo: I'd put comment above this js and above button code in the server. I'd like to avoid situation when somebody change button code in server and js will suddenly fail

gogonzo: @Polkas what is the purpuse of not adding shinyjs?

Polkas: > pacs::pac_deps("shinyjs", description_v = T)
       Package Version
1    base64enc        
2        bslib   0.3.0
3       cachem        
4   commonmark     1.7
5       crayon        
6       digest   0.6.8
7     ellipsis        
8      fastmap   1.1.0
9  fontawesome   0.2.1
10          fs        
11        glue   1.3.2
12   htmltools   0.5.2
13      httpuv   1.5.2
14   jquerylib   0.1.3
15    jsonlite  0.9.16
16       later   1.0.0
17   lifecycle   0.2.0
18    magrittr        
19     memoise        
20        mime     0.3
21    promises   1.1.0
22          R6     2.0
23    rappdirs        
24        Rcpp   1.0.7
25       rlang   1.0.5
26        sass   0.4.0
27       shiny   1.0.0
28 sourcetools        
29       withr        
30      xtable

Polkas: To simply have less dependencies, and running the JavaScript on the server side is not a good thing.

Polkas: comments added

gogonzo (approved):
👍

PR #152, opened 2022-10-13
disable

Polkas:
closes coredev#400

a true disable class as the default one is not truly disable clicking

### DIFF
--- a/inst/css/custom.css
+++ b/inst/css/custom.css
@@ -64,3 +64,10 @@
   min-width: 55px;
   margin: 1px 1px;
 }
+
+button.disabled {
+  color: grey;
+  cursor: not-allowed;
+  pointer-events: none;
+  text-decoration: none;
+}


gogonzo (commented):
I don't agree introducing more custom js and css. What do you think about disabling the button on the server side - no js needed then
```r
 shiny::observeEvent(input$add_card_ok, once = TRUE, {
```
https://github.com/insightsengineering/teal.reporter/blob/ea50cc0762bef2fae70b381ba7ab8026ecf2be7f/R/AddCardModule.R#L127

Polkas:
> I don't agree introducing more custom js and css. What do you think about disabling the button on the server side - no js needed then
> 
> ```r
>  shiny::observeEvent(input$add_card_ok, once = TRUE, {
> ```
> 
> https://github.com/insightsengineering/teal.reporter/blob/ea50cc0762bef2fae70b381ba7ab8026ecf2be7f/R/AddCardModule.R#L127

Your proposition looks be be more clear, agree.

gogonzo:
@Polkas guess what hehe, `once = TRUE` makes it even more strict - one can't click add report again for this module. It means your initial js and css makes sense. Sorry for that

nikolas-burkoff:
Can you use `ignoreInit = TRUE` though?

Polkas:
Maybe we should leave it as it is if sb will click many times he will get what he wants;p

PR #155, opened 2022-10-21
Enable multiline comments in report previewer and focus on 'add card' comment box

### LINKED ISSUE #71: Multi-line comments and modal dismissing
From https://github.com/insightsengineering/coredev-tasks/issues/244

- [ ] From @npaszty when you click in the add card button could the cursor be active in the comment box?
- [ ] ~The add a card to report modal gets closed when you click elsewhere on the app - I don't think that is the right behaviour~
- [ ] If I do this:
image

I get this:
image

(Although I'm not sure people would agree, being able to add markdown there would be cool)

asbates:
This PR enables markdown rendering in the report previewer. Users can use markdown in report card comment boxes and it will display in the previewer. Note that markdown in comment boxes already renders correctly in the downloaded report. The change here ensures rendering in the previewer within an app.

Also sets the cursor to focus in the comment box when the add report card button is clicked.

Note the first comment in the issues shows a multiline comment not rendering correctly. However the syntax is off because for markdown, there needs to be a blank line between each comment line.

There are no tests for this PR because

1. for the cursor focus, there isn't really an easy way to test this.
2. testing report previewer rendering requires upgrading to `testthat` 3rd edition. It was discussed by the team that we shouldn't upgrade this repo without upgrading the others.

Fixes #71

Example usage.

```
library(shiny)
library(teal.reporter)
library(ggplot2)

ui <- fluidPage(
  # please, specify specific bootstrap version and theme
  theme = bslib::bs_theme(version = "4"),
  titlePanel(""),
  tabsetPanel(
    tabPanel(
      "main App",
      tags$br(),
      sidebarLayout(
        sidebarPanel(
          uiOutput("encoding")
        ),
        mainPanel(
          tabsetPanel(
            id = "tabs",
            tabPanel("Plot", plotOutput("dist_plot"))
          )
        )
      )
    ),
    ### REPORTER
    tabPanel(
      "Previewer",
      reporter_previewer_ui("prev")
    )
    ###
  )
)
server <- function(input, output, session) {
  output$encoding <- renderUI({
    tagList(
      ### REPORTER
      teal.reporter::simple_reporter_ui("simple_reporter"),
      ###
      if (input$tabs == "Plot") {
        sliderInput(
          "binwidth",
          "binwidth",
          min = 2,
          max = 10,
          value = 8
        )
      } else {
        NULL
      }
    )
  })
  plot <- reactive({
    req(input$binwidth)
    x <- mtcars$mpg
    ggplot2::ggplot(data = mtcars, ggplot2::aes(x = mpg)) +
      ggplot2::geom_histogram(binwidth = input$binwidth)
  })
  output$dist_plot <- renderPlot(plot())
  
  ### REPORTER
  reporter <- teal.reporter::Reporter$new()
  card_fun <- function(card = ReportCard$new(), comment) {
    if (input$tabs == "Plot") {
      card$set_name("Plot Module")
      card$append_text("My plot", "header2")
      card$append_plot(plot())
      card$append_text(
        paste(
          c(
            "x <- mtcars$mpg",
            "ggplot2::ggplot(data = mtcars, ggplot2::aes(x = mpg)) +",
            paste0("ggplot2::geom_histogram(binwidth = ", input$binwidth, ")")
          ),
          collapse = "\n"
        ),
        "verbatim"
      )
    }
    if (!comment == "") {
      card$append_text("Comment", "header3")
      card$append_text(comment)
    }
    card
  }
  teal.reporter::simple_reporter_srv("simple_reporter", reporter = reporter, card_fun = card_fun)
  teal.reporter::reporter_previewer_srv("prev", reporter)
  ###
}

shinyApp(ui = ui, server = server)

```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ### New features
 
-* Updated the AddCardModule comment input to have an active cursor when adding a card.
+* Updated the `AddCardModule` comment input to have an active cursor when adding a card.
 * Updated report previewer to support preview of multiline comment.
 
 # teal.reporter 0.1.1
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@
 ### New features
 
 * Updated the AddCardModule comment input to have an active cursor when adding a card.
-* Updated report previewer to allow multiline comment.
+* Updated report previewer to support preview of multiline comment.
 
 # teal.reporter 0.1.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ### New features
 
-* Updated the comment block to have an active cursor when adding a card.
+* Updated the AddCardModule comment input to have an active cursor when adding a card.
 * Updated report previewer to allow multiline comment.
 
 # teal.reporter 0.1.1
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@
 ### New features
 
 * Updated the comment block to have an active cursor when adding a card.
-* Updated report previewer to render markdown as in the downloaded document.
+* Updated report previewer to allow multiline comment.
 
 # teal.reporter 0.1.1
 
--- a/tests/testthat/test-PreviewerReportModule.R
+++ b/tests/testthat/test-PreviewerReportModule.R
@@ -124,21 +124,4 @@ testthat::test_that("reporter_previewer_ui - returns a tagList", {
   )
 })
 
-testthat::test_that("render_text_block_preview - markdown renders to html fragment", {
-  block <- TextBlock$new()
 
-  block$set_content(
-    "
-  this is a multiline comment
-  line 2
-  line 3
-  "
-  )
-
-  block_html <- block_to_html(block)
-
-  testthat::expect_equal(
-    as.character(block_html),
-    "<pre>\n  this is a multiline comment\n  line 2\n  line 3\n  </pre>"
-  )
-})
--- a/tests/testthat/test-PreviewerReportModule.R
+++ b/tests/testthat/test-PreviewerReportModule.R
@@ -125,7 +125,6 @@ testthat::test_that("reporter_previewer_ui - returns a tagList", {
 })
 
 testthat::test_that("render_text_block_preview - markdown renders to html fragment", {
-
   block <- TextBlock$new()
 
   block$set_content(
--- a/tests/testthat/test-PreviewerReportModule.R
+++ b/tests/testthat/test-PreviewerReportModule.R
@@ -123,5 +123,3 @@ testthat::test_that("reporter_previewer_ui - returns a tagList", {
     inherits(reporter_previewer_ui("sth"), c("shiny.tag"))
   )
 })
-
-


NEWS.md
```diff
@@ -1,3 +1,10 @@
+# teal.reporter 0.1.1.9000
+
+### New features
+
+* When adding a card, the cursor is focused in the comment block.
+* Report Previewer renders markdown as in downloaded document.
```
mhallal1, line 6: According to our good practice, news entries are preferably in the following format:
"past verb + statement", for example:

mhallal1, line 6:
```suggestion
* Updated report previewer to render markdown as in the downloaded document.
```

NEWS.md
```diff
@@ -1,3 +1,10 @@
+# teal.reporter 0.1.1.9000
+
+### New features
+
+* When adding a card, the cursor is focused in the comment block.
```
mhallal1, line 5:
```suggestion
* Updated the comment block to have an active cursor when adding a card.
```

mhallal1:
@asbates  What about simple text with line breaks as in the original issue?

Polkas:
1. the render_text_block_preview function is overcomplicated
2. The jquery event Listener is to general about the element , checkout e.g. https://stackoverflow.com/questions/19279629/bootstrap-jquery-show-bs-modal-event-wont-fire
3. lack of added additional tests, for the markdown and cursor updates. Both of them seem to be able to check with shiny::testServer.

### Markdown

Example for the markdown::markdownToHTML function which convert markdown to HTML. THe `markdown` package is a RStudio tool and have only 2 dependencies. In my opinion the markdown package solution is more direct comapred to temp files and rmarkdown package.

```r
library(shiny)

ui <- fluidPage(
  textAreaInput("markdown", "markdown"),
  uiOutput("HTMLtext")
)

server <- function(input, output, session) {
  output$HTMLtext <- renderUI({
      shiny::HTML(markdown::markdownToHTML(text = input$markdown, fragment.only = TRUE))
  })
}

shinyApp(ui, server)
```

other solution cloud be to use JavaScript library like https://github.com/showdownjs/showdown directly

R/Previewer.R
```diff
@@ -181,6 +181,21 @@ block_to_html <- function(b) {
   )
 }
 
+#' @keywords internal
```
Polkas, line 184: I proposed a markdown or a pure Java Script solution.

asbates:
> @asbates What about simple text with line breaks as in the original issue?

For a test? Yes, that could work too. I'm hesitant about that as well because comparing to html in a string seems brittle to me. However, it can serve as a decent stop gap measure until we have snapshot testing enabled.

Polkas:
@asbates Please compare the markdown and rmarkdwon solutions with microbenchmark::microbenchmark
We have to assume each could be run x100 and more times each session.

R/AddCardModule.R
```diff
@@ -32,6 +32,21 @@ add_card_button_ui <- function(id) {
...
+            $(document).ready(function() {
+              $('body').on('shown.bs.modal', (e) => {
+                $(e.target).find('textarea').focus();
+              });
+            })
+            "
+          )
+        )
+      )
+    ),
```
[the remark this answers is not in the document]
asbates: I've updated this to be more specific. Instead of body, it uses `'#shiny-modal` which is the id Shiny uses for modals (bootstrap only allows 1 modal at a time). And instead of `event.target`, I've switched to an id, matching the text area input (`ns("comment")`)

asbates:
> > @asbates What about simple text with line breaks as in the original issue?
> 
> For a test? Yes, that could work too. I'm hesitant about that as well because comparing to html in a string seems brittle to me. However, it can serve as a decent stop gap measure until we have snapshot testing enabled.

@mhallal1 I've added a simple test case. I can add more and/or more complex cases if needed. But I really think snapshot testing is the way to go here, not checking HTML as a string.

R/Previewer.R
asbates: I don't think using JS is a great idea so of the two proposed options I'd go with markdown.

However, I don't see how it's worth it to add markdown when we already depend on rmarkdown. It works great in the simple example app you provided. But to feed input directly to `markdown::markdownToHTML` would require refactoring a lot of the code base. The add card modal and the report previewer are separate modules and the comment box content right now is stored in a `TextBlock` class. I don't think it would be simple to get all this working.

I'm OK going down this route if that's what is decided but that goes beyond this issue and PR. I think we would need more discussion and more details fleshed out before continuing in this way.

Polkas:
Current solution is highly inefficient.

``` r
render_text_block_preview <- function(block_content) {
  file <- tempfile()
  on.exit(rm(file))
  # need at least a title to prevent warnings
  header <- "---\ntitle: 'title'\n---\n"
  writeLines(paste0(header, block_content), file)
  html <- rmarkdown::render(
    file,
    output_format = "html_fragment",
    quiet = TRUE,
    params = list(title = "")
  )
  shiny::HTML(paste(readLines(html), collapse = ""))
}

input_small = "Hello World!"
microbenchmark::microbenchmark(
  markdown::markdownToHTML(text = input_small, fragment.only = TRUE),
  render_text_block_preview(input_small),
  unit = "s"
)
#> Warning in microbenchmark::microbenchmark(markdown::markdownToHTML(text =
#> input_small, : less accurate nanosecond times to avoid potential integer
#> overflows
#> Unit: seconds
#>                                                                expr         min          lq         mean
#>  markdown::markdownToHTML(text = input_small, fragment.only = TRUE) 0.000036859 0.000044567 6.725435e-05
#>                             render_text_block_preview(input_small) 0.043106252 0.046016452 4.834861e-02
#>       median          uq        max neval
#> 0.000082041 0.000085772 0.00016974   100
#> 0.047061932 0.047592287 0.18996473   100
```

<sup>Created on 2022-10-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

donyunardi:
@Polkas @asbates 
Thank you for separating markdown feature in different issue.
We will need to check with PO to see if this is a desired feature.

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # teal.reporter 0.1.1.9000
 
+### New features
+
+* Updated the comment block to have an active cursor when adding a card.
+* Updated report previewer to render markdown as in the downloaded document.
```
donyunardi, line 6:
```suggestion
* Updated report previewer to allow multiline comment.
```

donyunardi (changes requested):

asbates:
@Polkas [Bootstrap](https://getbootstrap.com/docs/4.0/components/modal/) only allows one modal to be open at a time. 
image.  That's why we can use `'#shiny-modal'` as a selector. Although the id is repeated, the object is added and removed from the DOM so there is never more than 1 `'#shiny-modal'`. 

From what I can determine, when an element is removed from the DOM any attached event listeners are removed. There is a caveat that there can't be a JS reference to that element. We never create a variable (`el = $('#shiny-modal')` so as far as I can tell the event listeners are removed when the modal is dismissed. 

To be extra sure though, we can call `$('#shiny-modal').off(...)`.  If we call `.off()` immediately after our `.on()` call, there is not enough time for the modal to render and cursor to focus in the comment box. But, we can call `.off()` in a `setTimeout()` to allow for rendering time. A 5 second timeout does the trick. I've set the time out to a higher value so I can double check the listener gets destroyed.

What do you think about this `setTimeout()` approach?

tests/testthat/test-PreviewerReportModule.R
```diff
@@ -123,3 +123,22 @@ testthat::test_that("reporter_previewer_ui - returns a tagList", {
...
+  "
+  )
+
+  block_html <- block_to_html(block)
+
+  testthat::expect_equal(
+    as.character(block_html),
+    "<pre>\n  this is a multiline comment\n  line 2\n  line 3\n  </pre>"
+  )
+})
```
Polkas, line 144:
```suggestion


```

tests/testthat/test-PreviewerReportModule.R
```diff
@@ -123,3 +123,5 @@ testthat::test_that("reporter_previewer_ui - returns a tagList", {
     inherits(reporter_previewer_ui("sth"), c("shiny.tag"))
   )
 })
+
+
```
Polkas, line 127:
```suggestion

```

Polkas:
@donyunardi Yes, the issue separation was my proposition. 
I think the footer in the issue was a small trap, which suggests nice to have feature here.
Thank you @asbates for a great research regarding bootstrap modal, I will double check it out now.

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # teal.reporter 0.1.1.9000
 
+### New features
+
+* Updated the comment block to have an active cursor when adding a card.
```
Polkas, line 5:
```suggestion
* Updated the AddCardModule comment input to have an active cursor when adding a card.
```

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # teal.reporter 0.1.1.9000
 
+### New features
+
+* Updated the AddCardModule comment input to have an active cursor when adding a card.
+* Updated report previewer to allow multiline comment.
```
Polkas, line 6:
```suggestion
* Updated report previewer to support preview of multiline comment.
```

Polkas (approved):
LGTM 

Great work:)

donyunardi (approved):

Polkas:
@asbates please feel free to merge the PR now. Remember to edit commit messages, usually we leave one most important phrase and coauthors.

asbates:
> @asbates please feel free to merge the PR now. Remember to edit commit messages, usually we leave one most important phrase and coauthors.

I'm not sure what you mean in the second sentence here. Could we talk about it tomorrow after standup?

Polkas:
@asbates 

https://user-images.githubusercontent.com/10676545/199218562-316a1ca7-29a8-48c5-9944-d9671200990e.mov

PR #159, opened 2022-10-26
reporter around show r code

### LINKED ISSUE #85: Collapsible Show R Code in the Previewer
The Show R Code take a sufficient part of the Previewer View of a Card, especially if the code is complex.
If Show R Code field will be collapsible and by default hidden then it could look more elegant.

image

### LINKED ISSUE #131: Not display SRC in the final report by default
Is it possible to not display the R code in the generated final report by default (in the pdf, html, word and pptx files)? The reason is the reports will be delivered to stakeholders, who are not programmers and would only need the analysis results instead of the source code. The app developer will still be able to retrieve the code from the Rmd to trace back or turn on the SRC in the Rmd if needed, maybe something like

```
```{r, results='hide'}
print("<R code here>")
```

Polkas:
closes #85 
closes #131 

I join these two issues as they are connected with the same design decision, e.g. how to identify the r code text block.
150 lines net code added where 100 is a copy paste of panel_item from teal.widgets. panel_item is not exported.

- the `panel_item` which is the biggest code part here is a copy paste from teal.widgets, will work for all Bootstrap versions.
- control of show r code inclusion is done with params filed in the Rmd file and additional RcodeBlock. We have to identify what block is code in the server side and when render the document.
- The encoding in the previewer is a outputUI now, as we need to control when to add the input for show r code inclusion (checked with `any_rcode_fun` function). This decision has some pros and cons, now we are able to control download button directly but there is additional ui part rendered on the server side. More than that the `reporter_previewer_ui` function is simplified because of it, in my opinion depreciate of its arguments is not needed as they are optional and probably not yet used.
- I remove the subtitle for show r code element in the report, it simplify the process for scenarios with/without show r code

Still the general design is a target, so no additional ui input when no show r code is there 

https://user-images.githubusercontent.com/10676545/198226216-ca986795-15a4-4f47-9881-ff5fc72ceb14.mov


BLAZEWIM:
1. This warning keeps appearing after switching to report previewer when at least one card is added:

```
example(tm_g_patient_profile, package = 'teal.osprey')
shinyApp(app$ui, app$server)

Warning in file(con, "r") :
  file("") only supports open = "w+" and open = "w+b": using the former
```

2. Minor: maybe change label from "Include Show R Code" to "Include R Code"

R/Renderer.R
```diff
@@ -140,6 +140,13 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
       switch(text_style,
         "default" = block_content,
         "verbatim" = paste0("\n```\n", block_content, "\n```\n"),
+        "rcode" =
+          paste(
```
BLAZEWIM, line 144: Is there a reason for a paste over sprintf?

Polkas: As I remember sprintf is quicker than paste0, and surely sprintf is more elegant for longer text.

```
microbenchmark::microbenchmark(paste0("Hello", " World"), sprintf("Hello %s", "World"))
Unit: nanoseconds
                         expr min  lq   mean median  uq  max neval
    paste0("Hello", " World") 779 820 922.09    820 861 7667   100
 sprintf("Hello %s", "World") 287 328 387.04    328 369 3321   100
```

so the performance difference is minimal.

Polkas:
@BLAZEWIM thank you for your time. Truly there was an error with unhelpful message, the css file path was wrong. I just fix it:)

 Regarding "Include Show R Code" -> "Include R Code", I treat the "Show R Code" as a brand/product name.

R/Renderer.R
Polkas: ah I see now, I test here different solutions and this paste should be removed:)
This was the places I fought a lot. 
Good catch.

DESCRIPTION
```diff
@@ -18,6 +18,7 @@ Imports:
     bslib,
     checkmate,
     grid,
+    htmltools,
```
Polkas, line 21: It is already there (e.g. shiny) so it is not a new dependency

R/utils.R
```diff
@@ -9,3 +9,103 @@ get_bs_version <- function() {
     "3"
   }
 }
+
```
Polkas, line 12: already a code base

Polkas:
TODO 
share the encoding  ui between Download module and previewer
rcode style public

Polkas:
I will think in the morning if the separate Block for SHow R code make sense. 
Instead of having rcode textBlock style, to implement the new SHowRCodeBlock which could be specified with any number of r chunk params like echo.

NEWS.md
```diff
@@ -1,5 +1,16 @@
 # teal.reporter 0.1.1.9000
 
+### New features
+* Added the collapsible panel for the Show R Code in the previewer.
+* Added additional input when needed to decided if Show R Code should be part of the report.
+* Added the new `rcode` `TextBlock` style.
```
Polkas, line 6:
```suggestion
* Added the new `RcodeBlock` back.
```

Polkas:
I will split it to separate PRs.

PR #161, opened 2022-10-28
RcodeBlock

### LINKED ISSUE #160: R code in the reporter
I propose to create the new RcodeBlock class to handle rmarkdown r chunks . R chunks are a main part of rmarkdown document so they should be a part of regular ReportCard.

Polkas:
closes #160 

is a blocker for 
https://github.com/insightsengineering/teal.reporter/pull/162/files
https://github.com/insightsengineering/teal.reporter/pull/163/files

Why we need RcodeBlock ... as the r code chunks are the main element of any rmarkdown document:

```
```{r, eval=TRUE, echo=TRUE}
2+2
```

More than that we we can easily identify r code now which is needed for linked PRs.

RcodeBlock class will be applied on the level of the general ReportCard as it will complement already available rmarkdown functionalities.


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # teal.reporter 0.1.1.9000
 
+### New features
+* Added the new `RcodeBlock` block for a custom `rmarkdown` r chunk.
+
```
Polkas, line 5:
```suggestion
### New features
* Added the new `RcodeBlock` block for a custom `rmarkdown` r chunk.

### Miscellaneous
* Updated `append_src` method of `TealReportCard` to not add additional "R Code" Subtitle.

```

R/RcodeBlock.R
```diff
@@ -0,0 +1,87 @@
...
+    #' @param content (`character(1)` or `character(0)`) a string assigned to this `RcodeBlock`
+    #' @param ... any `rmarkdown` R chunk parameter and it value.
+    #'
+    #' @return `RcodeBlock`
+    #' @examples
+    #' block <- teal.reporter:::RcodeBlock$new()
+    #'
+    initialize = function(content = character(0), ...) {
+      super$set_content(content)
+      self$set_params(list(...))
```
nikolas-burkoff, line 20: I wonder if params should be moved up to ContentBlock as for example TableBlock could make use of params
https://github.com/insightsengineering/teal.reporter/blob/e3adbba61688df35e2ebca92efb9cd72956d16a6/R/Renderer.R#L157

Also as for now we're using the global knitr options, do we need to add the complexity of params for the blocks at this time?

nikolas-burkoff (commented):
Comment from me

R/ReportCard.R
```diff
@@ -60,6 +60,18 @@ ReportCard <- R6::R6Class( # nolint: object_name_linter.
...
     },
+    #' @description Appends a `rmarkdown` R chunk to this `ReportCard`.
+    #'
+    #' @param text (`character(0)` or `character(1)`) the text
+    #' @param ... any `rmarkdown` R chunk parameter and its value.
+    #' @return invisibly self
+    #' @examples
+    #' card <- ReportCard$new()$append_rcode("2+2", echo = FALSE)
+    #'
+    append_rcode = function(text, ...) {
```
asbates, line 71: I'm thinking `code` or `src` (like `TealReportCard`) would be more clear here.

R/ReportCard.R
```diff
@@ -60,6 +60,18 @@ ReportCard <- R6::R6Class( # nolint: object_name_linter.
       private$content <- append(private$content, TextBlock$new(text, style))
       invisible(self)
     },
+    #' @description Appends a `rmarkdown` R chunk to this `ReportCard`.
+    #'
+    #' @param text (`character(0)` or `character(1)`) the text
```
asbates, line 65: Can we make it more explicit this is the code?

asbates: See also comment regarding argument name.

R/ReportCard.R
Polkas: The src is a show r code NEST related shortcut where as here we enable to add any r code.
Why r code not a code as we clearly communicating only r code is supported. 
In r markdown the other languages like python could be easily used too, I think only python could be added in future and with low probability. Then better to have a detail named method here.

R/ReportCard.R
Polkas: more than that there is no dot at the end of the line;o

R/RcodeBlock.R
Polkas: Good Question. It is strong suggestion of a small refactor here.
Then we should put the RcodeBlock before Picture and Table ones.
We should not put the params to ContentBlock as they are meaningless for TextBlock.
But the the moment the pictures ale loaded with regular markdown in the renderer, we not introduce the r code chunk with knitr picture load yet.

                        -> TextBlock
ContentBlock -> RcodeBlock  -> PictrueBLock
                                                   -> TableBlock

Not obvious one we could talk about htat.

PR #170, opened 2022-11-21
support the toc

### LINKED ISSUE #130: Adding table of content to report
image

Dear core dev fam

I was wondering if it is possible to add a navigation bar to the report?

Polkas:
closes #130 

I propose a  comprehensive solution for toc where it could be TRUE or FALSE. The additional arguments like toc_float are not supported as require another layer of complexity. 

This work so well as the `as_yaml_auto` is automatically resolve the flat structure to nested  for us.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 * Updated the `AddCardModule` comment input to have an active cursor when adding a card.
 * Updated report previewer to support preview of multiline comment.
-* Added support for a table of content for reporter documents.
+* Added support for a table of contents for reporter documents.
 
 ### New features
 * Added the new `RcodeBlock` block for a custom `rmarkdown` r chunk.
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 
 * Updated the `AddCardModule` comment input to have an active cursor when adding a card.
 * Updated report previewer to support preview of multiline comment.
+* Added support for a table of content for reporter documents.
 
 ### New features
 * Added the new `RcodeBlock` block for a custom `rmarkdown` r chunk.
--- a/R/DownloadModule.R
+++ b/R/DownloadModule.R
@@ -245,7 +245,7 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, session) {
             selected = rmd_yaml_args$output
           )
         ),
-        toc = shiny::checkboxInput(session$ns("toc"), label = "table of content", value = rmd_yaml_args$toc)
+        toc = shiny::checkboxInput(session$ns("toc"), label = "Include Table of Contents", value = rmd_yaml_args$toc)
       )
     })
   )
--- a/R/DownloadModule.R
+++ b/R/DownloadModule.R
@@ -38,7 +38,7 @@ download_report_button_ui <- function(id) {
 #' `pdflatex` has to be installed to use the `"pdf_document"` output. If vector is named then the names
 #' will appear in the `UI`.
 #' @param rmd_yaml_args `named list` vector with `Rmd` `yaml` header fields and their default values.
-#' Default `list(author = "NEST", title = "Report", date = Sys.Date(), output = "html_document")`.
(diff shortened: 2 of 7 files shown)


R/DownloadModule.R
```diff
@@ -50,15 +50,16 @@ download_report_button_srv <- function(id,
...
+                                         toc = FALSE
                                        )) {
   checkmate::assert_class(reporter, "Reporter")
   checkmate::assert_subset(rmd_output, c(
     "html_document", "pdf_document",
     "powerpoint_presentation", "word_document"
   ))
   checkmate::assert_list(rmd_yaml_args, names = "named")
-  checkmate::assert_true(all(c("author", "title", "date", "output") %in% names(rmd_yaml_args)))
+  checkmate::assert_true(all(c("author", "title", "date", "output", "toc") %in% names(rmd_yaml_args)))
```
chlebowa, line 62: Does this check whether supplied names are legal or whether all mandatory features have been specified?

Polkas: now we demand all of them but in future we want to support a subset from them.

chlebowa: Then I suggest changing 
```checkmate::assert_list(rmd_yaml_args, names = "named")``` 
to 
```checkmate::assert_list(rmd_yaml_args, names = "named", any.missing = FALSE)``` 
and adding type assertions on elements of `rmd_yaml_args`.

Comment (Polkas) [reply R/DownloadModule.R:62]: good call but it is out of the scope of this issue. please create a new issue to check the type for each element of rmd_ymal_args.

**Comment by chlebowa** [comment]
How about automatically add `toc_float =TRUE` to `as_yaml_auto` only if `output = "html_document"`?

**Comment by Polkas** [comment]
> How about automatically add `toc_float =TRUE` to `as_yaml_auto` only if `output = "html_document"`?

For sure we will not add it to the `as_yaml_auto` function which is a general solution and convert a flat list to nested one.
The step you propose could be added step after but what if sb do not want to use `toc_float =TRUE`.
Then we need to add additional input for `toc_float =TRUE` which have to be visible only for html.

File: NEWS.md [hunk 2]
```diff
@@ -4,6 +4,7 @@
 
 * Updated the `AddCardModule` comment input to have an active cursor when adding a card.
 * Updated report previewer to support preview of multiline comment.
+* Added support for a table of content for reporter documents.
```
nikolas-burkoff, line 7:
```suggestion
* Added support for a table of contents for reporter documents.
```

R/DownloadModule.R
```diff
@@ -50,15 +50,16 @@ download_report_button_srv <- function(id,
                                        ),
                                        rmd_yaml_args = list(
                                          author = "NEST", title = "Report",
-                                         date = as.character(Sys.Date()), output = "html_document"
+                                         date = as.character(Sys.Date()), output = "html_document",
+                                         toc = FALSE
```
nikolas-burkoff, line 54: Should we assert_flag(toc) ? - also below

R/DownloadModule.R
```diff
@@ -243,7 +244,8 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, session) {
             choices = rmd_output,
             selected = rmd_yaml_args$output
           )
-        )
+        ),
+        toc = shiny::checkboxInput(session$ns("toc"), label = "table of content", value = rmd_yaml_args$toc)
```
nikolas-burkoff, line 248: So to be consistent with the SRC. It should say "Include Table of Contents" (note in Eng you need the "s" for Contents)

nikolas-burkoff (approved):
Looks fine to me.

A couple of small comments in- line. 
Also do you think we should put the "report options" inside say a panel "Additional Options" 
image

R/DownloadModule.R
Polkas, line 248:
```suggestion
        toc = shiny::checkboxInput(session$ns("toc"), label = "Include Table of Contents", value = rmd_yaml_args$toc)
```

PR #172, opened 2022-11-22
edit card name

### LINKED ISSUE #94: Editable Card names
When we get the Add Card Modal we should add there a Card Name field additionally to already asked comment.
This Title will be used only for the Preview purposes or should we use it for the Card title (Block) too?
This will help to distinguish many cards from the same module , now they could easily have the same name in the previewer.

mhallal1:
closes #94 

After discussion with team members @nikolas-burkoff and @Polkas, we decided the following:
1. the module developer can set a default card label in the card function (it can be unnamed also)
2. the module user can change the card label with the UI input
3. No support for passing the module name by default to the reporter cards till the teal refactor is accomplished as it requires passing the label argument to the server function of all modules.

### DIFF
--- a/R/AddCardModule.R
+++ b/R/AddCardModule.R
@@ -146,7 +146,6 @@ add_card_button_srv <- function(id, reporter, card_fun) {
         card_fun_args_nams <- names(formals(card_fun))
         has_card_arg <- "card" %in% card_fun_args_nams
         has_comment_arg <- "comment" %in% card_fun_args_nams
-        has_label_arg <- "label" %in% card_fun_args_nams
 
         arg_list <- list()
 
--- a/R/AddCardModule.R
+++ b/R/AddCardModule.R
@@ -79,7 +79,7 @@ add_card_button_ui <- function(id) {
 add_card_button_srv <- function(id, reporter, card_fun) {
   checkmate::assert_function(card_fun)
   checkmate::assert_class(reporter, "Reporter")
-  checkmate::assert_subset(names(formals(card_fun)), c("card", "label", "comment"), empty.ok = TRUE)
+  checkmate::assert_subset(names(formals(card_fun)), c("card", "comment"), empty.ok = TRUE)
 
   shiny::moduleServer(
     id,
@@ -154,10 +154,6 @@ add_card_button_srv <- function(id, reporter, card_fun) {
           arg_list <- c(arg_list, list(comment = input$comment))
         }
 
-        if (has_label_arg && length(input$label) > 0 && input$label != "") {
-          arg_list <- c(arg_list, list(label = input$label))
-        }
-
         if (has_card_arg) {
           # The default_card is defined here because formals() returns a pairedlist object
           # of formal parameter names and their default values. The values are missing
@@ -191,9 +187,7 @@ add_card_button_srv <- function(id, reporter, card_fun) {
             card$append_text(input$comment)
           }
 
-          if (!has_label_arg && input$label == "" && length(card$get_name()) == 0) {
-            card$set_name("")
-          } else if (!has_label_arg && length(input$label) > 0 && input$label != "") {
+          if (length(input$label) > 0 && input$label != "") {
             card$set_name(input$label)
           }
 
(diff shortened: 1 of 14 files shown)


R/AddCardModule.R
```diff
@@ -89,6 +89,13 @@ add_card_button_srv <- function(id, reporter, card_fun) {
         shiny::modalDialog(
           easyClose = TRUE,
           shiny::tags$h3("Add a Card to the Report"),
+          shiny::textAreaInput(
```
Polkas, line 92: Simple Text Input is enough it will be a one Phrase we have to think what to do with the long ones.

R/AddCardModule.R
```diff
@@ -89,7 +89,13 @@ add_card_button_srv <- function(id, reporter, card_fun) {
         shiny::modalDialog(
           easyClose = TRUE,
           shiny::tags$h3("Add a Card to the Report"),
-          shiny::tags$hr(),
+          shiny::textInput(
+            ns("label"),
+            "Card Name",
+            value = "",
```
nikolas-burkoff, line 95: It feels like the modules should be able to give a default to here?

mhallal1: this requires passing `label` as an argument to `add_card_button_srv` and `simple_reporter_srv`.

nikolas-burkoff: Yup that's definitely a better way of adding it - though we'd still need to pass the label into the module args - might be best to not do that right now though and wait until teal is reconfigured to make that not needed...

nikolas-burkoff: > Another way can be to not have any default for label then to have empty name if not specified.when card is added.

Or to have the name the module developer has set for it (rather than the app developer).

mhallal1: so the latest update is:
1) the module developer can set a default card label in the card function
2) the module user can change the card label with the UI input
3) No support for passing the module name by default to the reporter cards till the teal refactor is accomplished as it requires passing the label argument to the server function of all modules.

R/AddCardModule.R
```diff
@@ -89,7 +89,13 @@ add_card_button_srv <- function(id, reporter, card_fun) {
         shiny::modalDialog(
           easyClose = TRUE,
           shiny::tags$h3("Add a Card to the Report"),
-          shiny::tags$hr(),
+          shiny::textInput(
```
Polkas, line 92:
```suggestion
          shiny::tags$hr(),
          shiny::textInput(
```

R/AddCardModule.R
```diff
@@ -180,6 +186,11 @@ add_card_button_srv <- function(id, reporter, card_fun) {
             card$append_text("Comment", "header3")
             card$append_text(input$comment)
           }
+
+          if (length(input$label) > 0 && input$label != "") {
```
Polkas, line 190:
```suggestion
          if ((!is.null(input$label)) && input$label != "") {
```

R/AddCardModule.R
```diff
@@ -140,7 +146,7 @@ add_card_button_srv <- function(id, reporter, card_fun) {
         card_fun_args_nams <- names(formals(card_fun))
         has_card_arg <- "card" %in% card_fun_args_nams
         has_comment_arg <- "comment" %in% card_fun_args_nams
-
+        has_label_arg <- "label" %in% card_fun_args_nams
```
Polkas, line 149: It is not used now. I think when have a label without a default value it will fail to eval `card <- try(do.call(card_fun, arg_list))`.

mhallal1: removed.
and actually, it does not fail.

Polkas: not possible you test it as at the beginning of server function we have `checkmate::assert_subset(names(formals(card_fun)), c("card", "comment"), empty.ok = TRUE)`.:)

Polkas: Then we get expected by me error.

Polkas: Please add description of the new arg in the roxygen2 #' @param card_fun ...

Polkas (approved):
LGTM

R/AddCardModule.R
mhallal1: done.

PR #176, opened 2022-12-12
knitr::include_graphics

### LINKED ISSUE #165: Renderer - loading images with knitr not markdown (![]) - proper width and height
We find out that markdown `{width=,height=}` is not general enough and for html everything after comma is not recognized.
The solution will be to use the knitr `knitr::include_graphics(width=,height=, dpi =)` with rmarkdown options for e.g. width and height. 
When applied please refactor the classes structure that the PictureBlocka and TAbleBlock inherits form RcodeBlock so both could have custom knitr options like echo.

```
              -> TextBlock
ContentBlock -> RcodeBlock  -> PictrueBLock
                            -> TableBlock
```

linked to https://github.com/insightsengineering/teal.reporter/issues/149

Polkas:
closes #165

Replace `` way of adding images with `knitr::include_graphics` which should be more stable across different documents types.


R/Renderer.R
```diff
@@ -174,8 +174,17 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
     pictureBlock2md = function(block) {
       basename_pic <- basename(block$get_content())
       file.copy(block$get_content(), file.path(private$output_dir, basename_pic))
+      params <- c(
+        `out.width` = "'100%'",
+        `out.height` = "'100%'"
+      )
       title <- block$get_title()
-      sprintf("![%s](%s){width=%s, height=%s}", title, basename_pic, block$get_dim()[1], block$get_dim()[2]) # nolint
```
mhallal1, line 178: we can solve this issue by removing the `,` between `width` and `height` as a comma is not required to separate them, see example:
`description`

mhallal1 (approved):
LGTM

PR #177, opened 2022-12-13
init Archiver module

### LINKED ISSUE #81: Create and implement a shiny module for Archiver class
linked to https://github.com/insightsengineering/teal.reporter/pull/80

We want to create a shiny module for the Archiver class and implement it in the vignettes.
The issue assume the Archiver class have to be adjusted when any additional functionalities are needed.

Remember it has to be applied for Previewer and single button functionality at the same time. 
The best if the button functionality (like add card module) could be simply used in the Previewer directly.
I imagine something similar to Add Card module.
In future there will be possible many ways for storing the data please take it into account, now we will work with zip file with JSON and static files (png and Rds).

Polkas:
closes #81 

Simply run the Previewer app from the vignette and test it yourself.

TODO

- test with different bootstrap versions (3,4, and 5)
- input validation 
- separate shiny app - a new vignette 
- load back, a way  to validate the archiver zip
- UI design
- docs
- test the module, testServer
- Should we record the Archiver creator?
- ...

https://user-images.githubusercontent.com/10676545/207322857-0d1ade4e-2e3b-48c5-a8ef-82a4849ad9ac.mov


lcd2yyz:
Pretty cool feature, very nice work!

**Question:** When the JSON file is downloaded, what's the actual content? Is it just the code/setting or the actual visualization? Asking because what if the input datasets changed between the download/upload?
**Comment 1:** This feature make more sense to live in the Previewer tab, as it's restoring the contents to be shown in the Previewer.
Comment 2: It's not clear from just looking at the name/label "Archiver" what the functionality actually do. Perhaps should consider a more self-explanatory naming or adding description label to indicate it's saving/restoring the analysis cards?

Polkas:
There are many possible improvements to this feature.

0. Question. That is a good comment. We already talked about the validation of datasets used for creating each card. Without additional protection, we will risk using different data in the same report (END USER responsibility). The idea of a dataset hash for each report card was proposed, which is automatic but adds complexity. Then Reporter (singleton instance - unique one for the whole session) requires an additional method like set_hash which will be optionally specified by the app developer when the reporter instance is created (the main unique element shared across the session). 
I have an even more straightforward idea like Archiver ID_ARCHIVER (server function argument) set by the shiny app developer, and later each archiver zip file name will have this ID_ARCHIVER. It is simple but not perfect, as the responsibility is entirely on the developer of the shiny app. This functionality should be optional for developer UX when the END USER responsibility is enough for the assumptions. PS. the downloaded file is the zip file with one JSON file (Reporter representation) and static files for images and tables.
1. This is for sure to be done. I already added it to the list too. However, for TEAL I recommend the save button on the direct level, as the save functionality should be at hand for the end users. We can have four buttons on the level of each TEAL module, so then we will add only the save button to the three that already exist and change the title. In the general case, the functionality is modular so that somebody can build any setup.
2: @insightsengineering/nest-core-dev, your naming ideas are wanted here. I appreciate your help. In the past, we name it Storager/Storage; currently, it is Archiver/Archive.

R/SimpleArchiver.R
```diff
@@ -0,0 +1,56 @@
...
+#' @param id `character(1)` this `shiny` module's id.
+#' @param reporter [`Reporter`] instance.
+#' @return `shiny::moduleServer`
+#' @export
+simple_archiver_srv <- function(id, reporter) {
+  shiny::moduleServer(
+    id,
+    function(input, output, session) {
+      archiver_load_srv("archive_load_simple", reporter = reporter)
+      archiver_save_srv("archive_save_simple", reporter = reporter)
```
BLAZEWIM, line 53:
```suggestion
      archiver_load_srv("archive_load_simple", reporter = reporter)
```

R/SimpleArchiver.R
```diff
@@ -0,0 +1,56 @@
...
+  shiny::tagList(
+    shiny::singleton(
+      shiny::tags$head(shiny::includeCSS(system.file("css/custom.css", package = "teal.reporter")))
+    ),
+    shiny::tags$div(
+      class = "block mb-4 p-1",
+      shiny::tags$label(class = "text-primary block -ml-1", shiny::tags$strong("Archiver")),
+      shiny::tags$div(
+        class = "simple_reporter_container",
+        archiver_load_ui(ns("archive_load_simple")),
```
BLAZEWIM, line 30: Isn't it better to have SAVE button first?

chlebowa:
> **Question:** When the JSON file is downloaded, what's the actual content? Is it just the code/setting or the actual visualization? Asking because what if the input datasets changed between the download/upload?

Do datasets have IDs?

Polkas:
After some talks with @mhallal1 I want to leave a small comment.
@pawelru and the team decided  we want teal.reporter tool to be as general as possible, and this way it was designed.
The priority is to make it working for any shiny app, NOT a teal one.
Because of its generality later can be easily implemented in the teal app.
This generality profits in bigger target group as can be used by the whole shiny community.
More than that code is more clear and easier to maintain, because of this generality.
@lcd2yyz for sure can change this strategy.

Polkas:
format name ?
do we need anything more than JSON?

Polkas:
I remove the archiver class after our discussions. We think the YAML implementation where we use yaml file instead of json is not needed. Moreover we see a little probability of implementing Data Base (e.g. remote MySQL) solution with login and password so end users do not have to manage the zip files. @nikolas-burkoff pointed a security issues for such too.

Main difference in the code is only in 2 functions.

Without Archiver:

```r

#' @keywords internal
load_json_archiver <- function(reporter, zip_path, filename) {
  tmp_dir <- tempdir()
  output_dir <- file.path(tmp_dir, sprintf("archiver_load_%s", gsub("[.]", "", format(Sys.time(), "%Y%m%d%H%M%OS4"))))
  dir.create(path = output_dir)
  if (!is.null(zip_path) && grepl("archiver_", filename)) {
    tryCatch(
      expr = zip::unzip(zip_path, exdir = output_dir, junkpaths = TRUE),
      warning = function(cond) {
        shiny::showNotification(
          ui = "Unzipping folder warning!",
          action = "Please contact app developer",
          type = "warning"
        )
      },
      error = function(cond) {
        shiny::showNotification(
          ui = "Unzipping folder error!",
          action = "Please contact app developer",
          type = "error"
        )
      }
    )
    reporter$from_jsondir(output_dir)
  } else {
    shiny::showNotification("Failed to load the archiver file.", type = "error")
  }
}

#' @keywords internal
archiver_download_handler_engine <- function(reporter, type, file) {
  switch(type,
    JSON = {
      tmp_dir <- tempdir()
      output_dir <- file.path(
        tmp_dir,
        sprintf("archiver_%s", gsub("[.]", "", format(Sys.time(), "%Y%m%d%H%M%OS4")))
      )
      dir.create(path = output_dir)
      archiver_dir <- reporter$to_jsondir(output_dir)
      temp_zip_file <- tempfile(fileext = ".zip")
      tryCatch(
        expr = zip::zipr(temp_zip_file, archiver_dir),
        warning = function(cond) {
          shiny::showNotification(
            ui = "Zipping folder warning!",
            action = "Please contact app developer",
            type = "warning"
          )
        },
        error = function(cond) {
          shiny::showNotification(
            ui = "Zipping folder error!",
            action = "Please contact app developer",
            type = "error"
          )
        }
      )

      tryCatch(
        expr = file.copy(temp_zip_file, file),
        warning = function(cond) {
          shiny::showNotification(
            ui = "Copying file warning!",
            action = "Please contact app developer",
            type = "warning"
          )
        },
        error = function(cond) {
          shiny::showNotification(
            ui = "Copying file error!",
            action = "Please contact app developer",
            type = "error"
          )
        }
      )
    }
  )
}
```

With Archiver:

```r

#' @keywords internal
load_json_archiver <- function(reporter, zip_path, filename) {
  archiver <- JSONArchiver$new()
  tmp_dir <- tempdir()
  output_dir <- file.path(tmp_dir, sprintf("archive_%s", gsub("[.]", "", format(Sys.time(), "%Y%m%d%H%M%OS4"))))
  dir.create(output_dir)
  tryCatch(
    expr = zip::unzip(zip_path, exdir = output_dir, junkpaths = TRUE),
    warning = function(cond) {
      shiny::showNotification(
        ui = "Unzipping folder warning!",
        action = "Please contact app developer",
        type = "warning"
      )
    },
    error = function(cond) {
      shiny::showNotification(
        ui = "Unzipping folder error!",
        action = "Please contact app developer",
        type = "error"
      )
    }
  )
  archiver$read(output_dir)
}

#' @keywords internal
archiver_download_handler_engine <- function(reporter, type, file) {
  switch(type,
    JSON = {
      tmp_dir <- tempdir()
      output_dir <- file.path(
        tmp_dir,
        sprintf("archiver_%s", gsub("[.]", "", format(Sys.time(), "%Y%m%d%H%M%OS4")))
      )
      dir.create(path = output_dir)
      archiver_dir <- reporter$to_jsondir(output_dir)
      temp_zip_file <- tempfile(fileext = ".zip")
      tryCatch(
        expr = zip::zipr(temp_zip_file, archiver_dir),
        warning = function(cond) {
          shiny::showNotification(
            ui = "Zipping folder warning!",
            action = "Please contact app developer",
            type = "warning"
          )
        },
        error = function(cond) {
          shiny::showNotification(
            ui = "Zipping folder error!",
            action = "Please contact app developer",
            type = "error"
          )
        }
      )

      tryCatch(
        expr = file.copy(temp_zip_file, file),
        warning = function(cond) {
          shiny::showNotification(
            ui = "Copying file warning!",
            action = "Please contact app developer",
            type = "warning"
          )
        },
        error = function(cond) {
          shiny::showNotification(
            ui = "Copying file error!",
            action = "Please contact app developer",
            type = "error"
          )
        }
      )
    }
  )
}
```

Polkas:
I can recall essential things now. 

The `from_jsondir` and `to_jsondir` seems to be suited for Archiver (removed in this PR) class, not Reporter. There were two reasons It was decided this way. As the whole process assumes a singleton design for the Reporter, the method on the level of the Reporter makes it very easy to update it. On the other hand, when these two methods were in Archiver, we needed the additional `from_reporter` method, and the process was less optimal. 
Another thing is that I was thinking of the Data Base solution based on a solution like AWS S3, then we could use the `from_jsondir` and `to_jsondir` methods for it too. We can send a directory directly to AWS S3 buckets. What class will manage such DB functionality? We need some additional encapsulation, then. 

To sum up, the whole problem of perceiving this item is how we think about other solutions for saving the report. Why the current could be better as the end user could end up with hundreds of files in his local Downloads folder. More than that, the end user is responsible for managing the files himself.

gogonzo:
Closing along with #81

PR #178, opened 2022-12-21
init the new modal for removal of a card in the previewer

### LINKED ISSUE #138: Inconsistent confirmation dialogs
I assume for backlog?

When deleting the whole report you get one type of dialog:

image

When deleting a single card you get a different alert:

image

Can these be the same style?

Polkas:
closes #138 

The remove card in the reporter from previewer functionality is partly transferred to the server side.
With this PR the modals should be consistent as are using the same shiny internals.
The thing we should validate is proposed event path =  jquery observer -> open modal observer -> remove card and modal observer.

### DIFF
--- a/R/Previewer.R
+++ b/R/Previewer.R
@@ -124,9 +124,14 @@ reporter_previewer_srv <- function(id,
       shiny::observeEvent(input$card_remove_id, {
         shiny::showModal(
           shiny::modalDialog(
-            title = sprintf(
-              "Do you really want to remove the card %s from the Report?",
-              input$card_remove_id
+            title = "Remove the Report Card",
+            shiny::tags$p(
+              shiny::HTML(
+                sprintf(
+                  "Do you really want to remove <strong>the card %s</strong> from the Report?",
+                  input$card_remove_id
+                )
+              )
             ),
             footer = shiny::tagList(
               shiny::modalButton("Cancel"),
--- a/R/Previewer.R
+++ b/R/Previewer.R
@@ -134,7 +134,14 @@ reporter_previewer_srv <- function(id,
               )
             ),
             footer = shiny::tagList(
-              shiny::modalButton("Cancel"),
+              shiny::tags$button(
+                type = "button",
+                class = "btn btn-danger",
+                `data-dismiss` = "modal",
+                `data-bs-dismiss` = "modal",
+                NULL,
+                "Cancel"
+              ),
               shiny::actionButton(ns("remove_card_ok"), "OK", class = "btn-warning")
             )
           )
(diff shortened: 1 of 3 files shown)


R/Previewer.R
```diff
@@ -122,7 +122,28 @@ reporter_previewer_srv <- function(id,
       })
 
       shiny::observeEvent(input$card_remove_id, {
+        shiny::showModal(
+          shiny::modalDialog(
+            title = "Remove the Report Card",
+            shiny::tags$p(
+              shiny::HTML(
+                sprintf(
+                  "Do you really want to remove <strong>the card %s</strong> from the Report?",
```
chlebowa, line 131:
```suggestion
                  "Do you really want to remove <strong>Card %s</strong> from the Report?",
```

chlebowa (changes requested):
Looks great.

Could you make the buttons consistent in all modals? Currently "Cancel" is red in old ones and grey in the new one. Likewise, the Reset button presents "Cancel" and "Reset", while the Remove button presents "Cancel" and "OK".

I also think red should be assigned to buttons that remove content, not abort the action. I was going to mention this in the next round of UAT but this seems like a good opportunity.

Polkas:
@chlebowa I agree with you that red color for cancel button looks to not be suited.
I created a new issue https://github.com/insightsengineering/teal.reporter/issues/179 . Do you think we should solved it in this PR.

R/Previewer.R
```diff
@@ -122,7 +122,28 @@ reporter_previewer_srv <- function(id,
...
+                  input$card_remove_id
+                )
+              )
+            ),
+            footer = shiny::tagList(
+              shiny::modalButton("Cancel"),
+              shiny::actionButton(ns("remove_card_ok"), "OK", class = "btn-warning")
```
asbates, line 138: I think we should use the same type of buttons as in the [report reset modal](https://github.com/insightsengineering/teal.reporter/blob/1649d79db3f515efca0fa9db79b74fc867022962/R/ResetModule.R#L64-L72).

Otherwise the two modals will be more similar than before but will still be inconsistent.

asbates (changes requested):
Looks good!

I think we can get them looking more similar by having the same buttons.

R/Previewer.R
Polkas: We are waiting for the response and preference from @lcd2yyz now, if we should update the colors in this PR. This PR was blocked because of the suggestion to not use the old colors you suggested to apply.

Polkas:
I updated the cancel button to btn-danger style to be consistent with the current design. 
I recommend to merge it and take care of the proposition for new colors for buttons in the new issue https://github.com/insightsengineering/teal.reporter/issues/179.

asbates (approved):

PR #186, opened 2023-05-11
Change color of buttons in modals

### LINKED ISSUE #179: cancel button color in a modal
linked to https://github.com/insightsengineering/teal.reporter/pull/178

Proposed by @chlebowa.
The red color for cancel buttons seems not not be suited. Why stopping the action is dangerous, I think grey will be more suited. On the other hand a removal button should be red.

image

asbates:
Changes the colors of buttons in the various modals.

- "Cancel" buttons are color "secondary".
- "Remove" buttons are color "danger"
- "Add" buttons are color "primary"

Fixes #179

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@
 
 ### Enhancements
 
+* Changed modal button colors to align with their purpose and with other modals.
 * Updated the `AddCardModule` comment input to have an active cursor when adding a card.
 * Updated report previewer to support preview of multiline comment.
 * Added support for a table of contents for reporter documents.
--- a/R/AddCardModule.R
+++ b/R/AddCardModule.R
@@ -123,7 +123,7 @@ add_card_button_srv <- function(id, reporter, card_fun) {
           footer = shiny::div(
             shiny::tags$button(
               type = "button",
-              class = "btn btn-danger",
+              class = "btn btn-secondary",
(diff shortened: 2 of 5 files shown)


donyunardi:
Felt like we can do better when the button is disabled (blue and grey such a weird combination)

image

image

How about we make these changes  in `Previewer.css`, 
https://github.com/insightsengineering/teal.reporter/blob/67c71ea6ac00982676792c58942af4c69cfc3d52/inst/css/Previewer.css#L10-L15

To be like this:
```r
a.disabled {
  color: white;
  background-color:#6a98bf;
  border-color: #6a98bf;
  cursor: not-allowed;
  pointer-events: none;
  text-decoration: none;
}
```

So now, the button looks like this when disabled:
image

And this when enabled:
image

donyunardi:
I just read my comment again and realized that this might be out of scope from the main issue.
I'd be happy to create a new issue if we can't squeeze this in.

asbates:
@donyunardi I think the disabled color should be a new issue. Mostly because it needs more thought around how to adjust the color. While your suggested style changes look better, I wouldn't recommend using a color directly like #6a98bf. It match the current theme, but what if I use Bootswatch United? What if I define my own custom primary color?

kartikeyakirar (approved):
Tested changes for issues. PR looks good.

PR #191, opened 2023-07-12
Refactor of the filter-panel

gogonzo:
part of https://github.com/insightsengineering/teal.slice/pull/165

### DIFF
--- a/R/TealReportCard.R
+++ b/R/TealReportCard.R
@@ -38,10 +38,6 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
     #'
     #' @param fs (`list`) of filter states.
     #' @return invisibly self
-    #' @examples
-    #' card <- TealReportCard$new()$append_fs(list(a = 1, b = 2))
-    #' card$get_content()[[1]]$get_content()
-    #'
     append_fs = function(fs) {
       # todo: whole class should be moved to teal
       checkmate::assert_class(fs, "teal_slices")
@@ -56,8 +52,8 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
             x_list$range <- paste(x_list$selected, collapse = " - ")
             x_list["selected"] <- NULL
           }
-          if (!is.null(x$arg)) {
-            x_list$arg <- if (x$arg == "subset") "Genes" else "Samples"
+          if (!is.null(x_list$arg)) {
+            x_list$arg <- if (x_list$arg == "subset") "Genes" else "Samples"
           }
 
           x_list <- x_list[
--- a/R/TealReportCard.R
+++ b/R/TealReportCard.R
@@ -43,11 +43,34 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
     #' card$get_content()[[1]]$get_content()
     #'
     append_fs = function(fs) {
+      # todo: whole class should be moved to teal
       checkmate::assert_class(fs, "teal_slices")
-
       if (length(fs) != 0) {
+        states_list <- lapply(fs, function(x) {
+          x_list <- reactiveValuesToList(x)
+          if (
+            inherits(x_list$choices, c("integer", "numeric", "Date", "POSIXct", "POSIXlt")) &&
+            length(x_list$choices) == 2 &&
+            length(x_list$selected) == 2
+          ) {
+            x_list$range <- paste(x_list$selected, collapse = " - ")
+            x_list["selected"] <- NULL
+          }
+          if (!is.null(x$arg)) {
+            x_list$arg <- if (x$arg == "subset") "Genes" else "Samples"
+          }
+
+          x_list <- x_list[
+            c("dataname", "varname", "experiment", "arg", "expr", "selected", "range", "keep_na", "keep_inf")
+          ]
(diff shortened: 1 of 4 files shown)


R/TealReportCard.R
```diff
@@ -43,21 +43,36 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
...
     append_fs = function(fs) {
-      checkmate::assert_list(fs)
-      attr_fs <- attr(fs, "formatted")
-      checkmate::assert_character(attr_fs, null.ok = TRUE)
-
+      # todo: whole class should be moved to teal
+      checkmate::assert_class(fs, "teal_slices")
       if (length(fs) != 0) {
+        states_list <- lapply(fs, function(x) {
+          x_list <- shiny::isolate(shiny::reactiveValuesToList(x))
```
chlebowa, line 50:
```suggestion
          x_list <- as.list(x)
```

R/TealReportCard.R
```diff
@@ -43,21 +43,36 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
...
+          if (
+            inherits(x_list$choices, c("integer", "numeric", "Date", "POSIXct", "POSIXlt")) &&
+              length(x_list$choices) == 2 &&
+              length(x_list$selected) == 2
+          ) {
+            x_list$range <- paste(x_list$selected, collapse = " - ")
+            x_list["selected"] <- NULL
+          }
+          if (!is.null(x$arg)) {
+            x_list$arg <- if (x$arg == "subset") "Genes" else "Samples"
```
chlebowa, line 60:
```suggestion
            x_list$arg <- if (x_list$arg == "subset") "Genes" else "Samples"
```

R/TealReportCard.R
```diff
@@ -43,21 +43,36 @@ TealReportCard <- R6::R6Class( # nolint: object_name_linter.
...
-      checkmate::assert_character(attr_fs, null.ok = TRUE)
-
+      # todo: whole class should be moved to teal
+      checkmate::assert_class(fs, "teal_slices")
       if (length(fs) != 0) {
+        states_list <- lapply(fs, function(x) {
+          x_list <- shiny::isolate(shiny::reactiveValuesToList(x))
+          if (
+            inherits(x_list$choices, c("integer", "numeric", "Date", "POSIXct", "POSIXlt")) &&
+              length(x_list$choices) == 2 &&
```
chlebowa, line 53: When `choices` is of length 2, range-like variables are converted to factor and passed to `ChoicesFilterState`, so selected values should be enumerated.

chlebowa (changes requested):
Ranges still come out as choices and choices are printed one per row.

```
append_fs = function(fs) {
  # todo: whole class should be moved to teal
  checkmate::assert_class(fs, "teal_slices")
  if (length(fs) != 0) {
    states_list <- lapply(fs, function(x) {
      x_list <- as.list(x)
      if (
        inherits(x_list$choices, c("integer", "numeric", "Date", "POSIXct", "POSIXlt")) &&
        length(x_list$choices) == 2 &&
        length(x_list$selected) == 2
      ) {
        x_list$range <- paste(x_list$selected, collapse = " - ")
        x_list["selected"] <- NULL
      }
      if (!is.null(x_list$arg)) {
        x_list$arg <- if (x$arg == "subset") "Genes" else "Samples"
      }
      
      x_list <- x_list[
        c("dataname", "varname", "experiment", "arg", "expr", "selected", "range", "keep_na", "keep_inf")
      ]
      names(x_list) <- c(
        "Dataset name", "Variable name", "Experiment", "Filtering by", "Applied expression",
        "Selected Values", "Selected range", "Include NA values", "Include Inf values"
      )
      
      Filter(Negate(is.null), x_list)
    })
    
    # self$append_text("Filter State", "header3")
    # self$append_text(yaml::as.yaml(states_list), "verbatim")
  }
  # self$append_metadata("FS", fs)
  # invisible(self)
  yaml::as.yaml(states_list)
}
x <- teal_slices(
  teal_slice("iris", "Species", selected = c("virginica", "setosa"), keep_na = TRUE),
  teal_slice("iris", "Sepal.Length", selected = c(4, 6), keep_inf = TRUE),
  count_type = "all"
)
append_fs(x) %>% cat
```

R/TealReportCard.R
gogonzo: This package shouldn't depend on `teal.slice` so `as.list` shouldn't be used neither. After we merge it I'm planning to move `teal.reporter::TealReporterCard` to follow a proper dependency direction. 
- `teal` imports `teal.reporter` so `teal` should extend `teal.reporter` features not otherwise.


I'm postponing this move as we will have to open `filter_panel_refactor@main` branch also in `teal.modules.*` and replace `teal.reporter::TealReporterCard` in all modules.

R/TealReportCard.R
gogonzo: Yes, problem is that we have no information available which can help us deduct this scenario. We don't know the class of `FilterState`, just `teal_slice`. We don't know original `x` values. Also `"getOption("teal.threshold_slider_vs_checkboxgroup")`, doesn't say much.

R/TealReportCard.R
chlebowa: All right.

chlebowa (approved):

PR #192, opened 2023-07-17
Move TealReportCard to teal

gogonzo:
Closes https://github.com/insightsengineering/teal/issues/864

- Moved `TealReportCard` to `teal`
- Changed code of the classes to not being dependent on `switch(class(obj)[1], ...)`. This factory design causing some problems as it is impossible to extend any of `teal.reporter` classes since method dispatch is hardcoded in this package. `inherits` solves some of the problems but not entirely. In the future we might consider moving some methods to the `*Block` classes but it needs to be preceeded with some design analysis (what should decide about output, `Renderer`s `Previewers` or `*ReportCard`, `*Block`. We should revisit this later)


chlebowa (approved):
Looks OK.

> `inherits` solves some of the problems but not entirely. 

I suppose some S3 dispatch would be in order?

Issue #169, opened 2022-11-19
Automatically wrap up the long code in the Report


Polkas:
@nikolas-burkoff I think I found the solution to automatically wrap up the R code in the Report.
We need the formatR package and additional knitr global option. We could easily add this option as now we support global knitr ones.

`knitr::opts_chunk$set(tidy.opts = list(width.cutoff = 60), tidy = TRUE)`

or tidy = "styler"

Source: https://stackoverflow.com/a/74502546/5442527

averissimo:
Optimal line length: https://baymard.com/blog/line-length-readability

chlebowa:
Acceptance criteria:

Figure out how to neatly wrap code in all report formats available.

m7pr:
Should we change global knitr options in here?
https://github.com/insightsengineering/teal.reporter/blob/main/R/Renderer.R#L38

kartikeyakirar:
> knitr::opts_chunk$set(tidy.opts = list(width.cutoff = 60), tidy = TRUE)

This has dependencies for formatR, but the formatting looks good for other formats. Should we add it for all formats or just for PDF?

m7pr:
nah, just put it in the default global_knitr options and this will be applied to every format

PR #218, opened 2023-09-26
169 wrapping up the long code in the Report

kartikeyakirar:
this Fixes https://github.com/insightsengineering/teal.reporter/issues/169

 I've implemented a solution that relies on the formatR package. I've ensured that this solution checks for the package in the namespace; if it's not accessible, the user will be notified. To maintain flexibility, I've included the formatR package in the suggested dependencies, as there are no strict dependencies on it.

In response to the feedback provided by @m7pr in [this comment](https://github.com/insightsengineering/teal.reporter/issues/169#issuecomment-1733269638), I've updated the default value of global_knitr to list(echo = TRUE, tidy.opts = list(width.cutoff = 60), tidy = FALSE).

Please note that there was an alternative option to set tidy = "styler", but I opted against it due to the additional dependency it would introduce (the styler package). Feel free to review the changes and let me know if any further adjustments are needed.

----
This branch utilizes the functionality provided by the 'formatR' package to automatically add spaces and indentations to the code. However, it does create a strong dependency on the 'formatR' package. If the package is not installed, the output will be generated without any formatting.

Additionally, this branch introduces the following changes:

- It exposes Knitr options to users through the functions 'download_report_button_srv', 'reporter_previewer_srv,' 'simple_reporter_srv,' and the R6 renderer.
- The Roxygen documentation has been updated to reflect these modifications.

changes in reports are observed as follows.

branch: main
image

branch: 169_wrap_rcode@main
image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,6 +31,7 @@ Imports:
     zip (>= 1.1.0)
 Suggests:
     DT (>= 0.13),
+    formatR (>= 1.5),
     ggplot2 (>= 3.4.0),
     lattice (>= 0.18-4),
     png,
@@ -44,8 +45,8 @@ RdMacros:
 Config/Needs/verdepcheck: rstudio/bslib, mllg/checkmate,
     rstudio/htmltools, yihui/knitr, r-lib/lifecycle, r-lib/R6,
     rstudio/rmarkdown, rstudio/shiny, dreamRs/shinyWidgets,
-    yaml=vubiostat/r-yaml, r-lib/zip, davidgohel/flextable, rstudio/DT,
-    tidyverse/ggplot2, deepayan/lattice, cran/png,
+    yaml=vubiostat/r-yaml, r-lib/zip, davidgohel/flextable, rstudio/DT, 
+    yihui/formatR, tidyverse/ggplot2, deepayan/lattice, cran/png,
     insightsengineering/rtables, r-lib/testthat, rstudio/tinytex
 Config/Needs/website: insightsengineering/nesttemplate
 Encoding: UTF-8
--- a/R/DownloadModule.R
+++ b/R/DownloadModule.R
@@ -33,11 +33,16 @@ download_report_button_ui <- function(id) {
 #' For more details see the vignette: `vignette("simpleReporter", "teal.reporter")`.
 #' @param id `character(1)` this `shiny` module's id.
 #' @param reporter [`Reporter`] instance.
+#' @param global_knitr `list` a of `knitr` parameters (passed to `knitr::opts_chunk$set`)
+#'  for customizing the rendering process.
 #' @inheritParams reporter_download_inputs
 #' @return `shiny::moduleServer`
+#' @details `r global_knitr_details()`
+#'
 #' @export
 download_report_button_srv <- function(id,
                                        reporter,
+                                       global_knitr = getOption("teal.reporter.global_knitr"),
                                        rmd_output = c(
                                          "html" = "html_document", "pdf" = "pdf_document",
                                          "powerpoint" = "powerpoint_presentation", "word" = "word_document"
@@ -48,6 +53,7 @@ download_report_button_srv <- function(id,
                                          toc = FALSE
                                        )) {
   checkmate::assert_class(reporter, "Reporter")
+  checkmate::assert_subset(names(global_knitr), names(knitr::opts_chunk$get()))
   checkmate::assert_subset(
     rmd_output,
     c(
@@ -134,8 +140,7 @@ download_report_button_srv <- function(id,
           shiny::showNotification("Rendering and Downloading the document.")
           input_list <- lapply(names(rmd_yaml_args), function(x) input[[x]])
           names(input_list) <- names(rmd_yaml_args)
-          global_knitr <- list()
-          if (is.logical(input$showrcode)) global_knitr <- list(echo = input$showrcode)
+          if (is.logical(input$showrcode)) global_knitr[["echo"]] <- input$showrcode
           report_render_and_compress(reporter, input_list, global_knitr, file)
         },
         contentType = "application/zip"
--- a/R/Previewer.R
+++ b/R/Previewer.R
@@ -31,10 +31,15 @@ reporter_previewer_ui <- function(id) {
 #' For more details see the vignette: `vignette("previewerReporter", "teal.reporter")`.
 #' @param id `character(1)` this `shiny` module's id.
 #' @param reporter `Reporter` instance
+#' @param global_knitr `list` a of `knitr` parameters (passed to `knitr::opts_chunk$set`)
+#'  for customizing the rendering process.
 #' @inheritParams reporter_download_inputs
+#' @details `r global_knitr_details()`
+#'
 #' @export
 reporter_previewer_srv <- function(id,
                                    reporter,
+                                   global_knitr = getOption("teal.reporter.global_knitr"),
                                    rmd_output = c(
                                      "html" = "html_document", "pdf" = "pdf_document",
                                      "powerpoint" = "powerpoint_presentation",
@@ -45,6 +50,7 @@ reporter_previewer_srv <- function(id,
                                      toc = FALSE
                                    )) {
   checkmate::assert_class(reporter, "Reporter")
+  checkmate::assert_subset(names(global_knitr), names(knitr::opts_chunk$get()))
   checkmate::assert_subset(
     rmd_output,
     c(
@@ -179,8 +185,7 @@ reporter_previewer_srv <- function(id,
           shiny::showNotification("Rendering and Downloading the document.")
           input_list <- lapply(names(rmd_yaml_args), function(x) input[[x]])
           names(input_list) <- names(rmd_yaml_args)
-          global_knitr <- list()
-          if (is.logical(input$showrcode)) global_knitr <- list(echo = input$showrcode)
+          if (is.logical(input$showrcode)) global_knitr[["echo"]] <- input$showrcode
           report_render_and_compress(reporter, input_list, global_knitr, file)
         },
         contentType = "application/zip"
--- a/R/Renderer.R
+++ b/R/Renderer.R
@@ -22,14 +22,16 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
     #' @description getting the `Rmd` text which could be easily rendered later.
     #'
     #' @param blocks `list` of `c("TextBlock", "PictureBlock", "NewpageBlock")` objects.
-    #' @param yaml_header `character` a `rmarkdown` `yaml` header.
-    #' @param global_knitr `list` a global `knitr` parameters, like echo.
-    #' But if local parameter is set it will have priority.
-    #' Defaults to empty `list()`.
+    #' @param yaml_header `character` an `rmarkdown` `yaml` header.
+    #' @param global_knitr `list` a of `knitr` parameters (passed to `knitr::opts_chunk$set`)
+    #'  for customizing the rendering process.
+    #' @details `r global_knitr_details()`
     #'
     #' @return `character` a `Rmd` text (`yaml` header + body), ready to be rendered.
-    renderRmd = function(blocks, yaml_header, global_knitr = list()) {
+    renderRmd = function(blocks, yaml_header, global_knitr = getOption("teal.reporter.global_knitr")) {
       checkmate::assert_list(blocks, c("TextBlock", "PictureBlock", "NewpageBlock", "TableBlock", "RcodeBlock"))
+      checkmate::assert_subset(names(global_knitr), names(knitr::opts_chunk$get()))
+
       if (missing(yaml_header)) {
         yaml_header <- md_header(yaml::as.yaml(list(title = "Report")))
       }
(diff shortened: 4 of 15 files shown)


m7pr:
@pawelru what are your thoughts on this? I know we had some conversation about packages being in Suggests and being checked on a code execution, but this is totally minor and I would accept this the way it is.

R/DownloadModule.R
```diff
@@ -136,6 +136,11 @@ download_report_button_srv <- function(id,
           names(input_list) <- names(rmd_yaml_args)
           global_knitr <- list()
           if (is.logical(input$showrcode)) global_knitr <- list(echo = input$showrcode)
+          if (requireNamespace("formatR", quietly = TRUE)) {
+            global_knitr <- append(global_knitr, list(tidy.opts = list(width.cutoff = 60), tidy = TRUE))
+          } else {
+            message("For better code formatting, consider installing the formatR package.")
+          }
```
m7pr, line 143: Hey @kartikeyakirar the `global_knitr` used here is hidden from the user, and only used when `download`ed.
There are other ways of creating our report, and this happens directly through `render_and_compress`
https://github.com/insightsengineering/teal.reporter/blob/69b6e7ce93e4fe1c6f4fa3245e4a587cc99a5b9b/R/DownloadModule.R#L161

This `global_knitr` is also used in Previewer and Renderer. https://github.com/search?q=repo%3Ainsightsengineering%2Fteal.reporter%20global_knitr&type=code

Let's unify the approach please. Can you change methods/functions that are exposed to a user ? https://github.com/insightsengineering/teal.reporter/issues/169#issuecomment-1733269638
Can you change the default values of all `global_knitr` parameters in all functions' declarations so it's visible something is setup by default, and not hidden inside the function?

m7pr (changes requested):

R/DownloadModule.R
kartikeyakirar: Got it. my intital impression was to fix the issues in the reports we downloaded. The renderRmd function generates the RMD file, and this adjustment can be made there.

> Can you change the default values of all global_knitr parameters in all functions' declarations so it's visible something is setup by default, and not hidden inside the function?

Yes, I can add the `global_knitr ` parameter to function declarations. However, I need to ensure that the user provides the correct parameter and that it's safely overwritten based on internal logic.

m7pr: jesus, this is the third time I need to say the same thing. I will actually show you with an arrow. Please change DEFAULT values of the function so it's exposed in the manual page of the function and visible for the user

image

do not hardcode this stuff in the middle of the function as it's not clear for the person running the function that some hidden behavior is used

image

kartikeyakirar: I changed it to draft mode while making updates. Sorry for confusion.

kartikeyakirar:
I have set the default options for global_knitr ` list(echo = TRUE, tidy.opts = list(width.cutoff = 60), tidy = FALSE)`
tidy.opts is set based on https://yihui.org/knitr/options/#chunk-options whereas tidy is FAlSE as default but will be changed to TRUE if formatR package is found in the namespace.

m7pr:
@kartikeyakirar what if we create 

```r
options(teal.reporter.knit.global = list(echo = TRUE, tidy.opts = list(width.cutoff = 60), tidy = FALSE))
```
on a package startup, and then we reuse `getOption('teal.reporter.knit.global')` in defaults of functions usage? This way we have code that is repeated in many places, but specified once, and is easy to be edited in package, and also during an R sessionl

m7pr:
https://r-pkgs.org/code.html#sec-code-onLoad-onAttach

kartikeyakirar:
> on a package startup, and then we reuse getOption('teal.reporter.knit.global') in defaults of functions usage? This way we have code that is repeated in many places, but specified once, and is easy to be edited in package, and also during an R sessionl

It's a good idea. 

FYI: I have removed the tidy.opt now. so it would be  `options(teal.reporter.knit.global = list(echo = TRUE,  tidy = FALSE) )`

m7pr:
please come up with a better name for `teal.reporter.knit.global` I think I mage a typo in it, and maybe it should be `global.knitr` instead of `knitr.global`

m7pr:
Even though `tidy = 'formatR` sets `tidy.opts = list(width.cutoff = 60)` I think it's worth it to pass `tidy.opts = list(width.cutoff = 60)` so it's clear which option is responsible for setting the width.

pawelru:
> @pawelru what are your thoughts on this? I know we had some conversation about packages being in Suggests and being checked on a code execution, but this is totally minor and I would accept this the way it is.

IMHO, a functionality-dependent package might be added to the Suggest (as opposed to Imports) only if a given package (with all of its features) will remain fully functional without this package installed. An anti-pattern is an error message requesting additional installs.

From what I can see, this hold true in this PR.

R/Renderer.R
```diff
@@ -23,17 +23,23 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
...
+      checkmate::assert_subset(names(global_knitr), names(knitr::opts_chunk$get()))
+
       if (missing(yaml_header)) {
         yaml_header <- md_header(yaml::as.yaml(list(title = "Report")))
       }
 
+      if (requireNamespace("formatR", quietly = TRUE)) {
+        global_knitr[["tidy"]] <- TRUE
+      } else {
+        message("For better code formatting, consider installing the formatR package.")
```
pawelru, line 40: does that mean that I will have a message on every render of report (or even worse: card)?

can we remove it?

kartikeyakirar: This message will be displayed in the console for every report rendering, except for the card. While we can remove it from the card, I believe it's important to inform the developer . IF we don't want to inform for every rendering then I can move the message to the .onLoad() function, ensuring it's displayed only once per user session.

kartikeyakirar: I see your point. I still think we should notify the app-developer as we are using formatR functionality However, I agree it shouldn't appear during rendering. Perhaps, during the .onload() event unless there are  reasons not to do so.

R/zzz.R
```diff
@@ -0,0 +1,22 @@
+.onLoad <- function(libname, pkgname) {
+  # adapted from https://r-pkgs.org/code.html#sec-code-onLoad-onAttach
+  op <- options()
+  default_global_knitr <- list(teal.reporter.global_knitr = list(
+    echo = TRUE,
+    tidy.opts = list(width.cutoff = 60),
+    tidy = FALSE
```
m7pr, line 7: @kartikeyakirar I think this should be `tidy = TRUE` and it will just not work if formattR is missing. do not hide stuff like that from the user https://github.com/insightsengineering/teal.reporter/pull/218/files#diff-c8b290bb66d11d029d9e7a7585d6aeb3628e578143c2fa1a44e4661636f26f7aR37-R39

R/zzz.R
```diff
@@ -0,0 +1,22 @@
...
+  ))
+
+  toset <- !(names(default_global_knitr) %in% names(op))
+  if (any(toset)) options(default_global_knitr[toset])
```
m7pr, line 11:
```suggestion
  if (toset) options(default_global_knitr)
```
@kartikeyakirar there is only one element in `default_global_knitr` list

R/zzz.R
kartikeyakirar: Initially, I set it to TRUE. However, if tidy = TRUE is  set, the report does not download because the knitr process is halted due to a warning in app.

R/zzz.R
chlebowa: If there is only one element, a clearer condition would be
`if ("teal.reporter.global_knitr" %in% names(options))`

R/zzz.R
kartikeyakirar: image

m7pr: @kartikeyakirar can you make it

```r
list(teal.reporter.global_knitr = list(
    echo = TRUE,
    tidy.opts = list(width.cutoff = 60),
    tidy = requireNamespace('formattR')
```

then?

So it's visible, and exposed to the end user and not hidden

m7pr (changes requested):
hey @kartikeyakirar can we point to what function this is passed? so that new and fresh users knows for which options to look up. Can you also extend `@description` or `@details` field with the list of default values in `getOption("teal.reporter.global_knitr")` ?

R/Previewer.R
```diff
@@ -31,10 +31,12 @@ reporter_previewer_ui <- function(id) {
 #' For more details see the vignette: `vignette("previewerReporter", "teal.reporter")`.
 #' @param id `character(1)` this `shiny` module's id.
 #' @param reporter `Reporter` instance
+#' @param global_knitr `list` a global `knitr` parameters for customizing the rendering process.
```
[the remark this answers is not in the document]
kartikeyakirar: I have added this as details for all the function using global_knitr parameter.

R/DownloadModule.R
```diff
@@ -33,9 +33,17 @@ download_report_button_ui <- function(id) {
 #' For more details see the vignette: `vignette("simpleReporter", "teal.reporter")`.
 #' @param id `character(1)` this `shiny` module's id.
 #' @param reporter [`Reporter`] instance.
-#' @param global_knitr `list` a global `knitr` parameters for customizing the rendering process.
+#' @param global_knitr `list` a of `knitr` parameters (passed to `knitr::opts_chunk$set`)
+#'  for customizing the rendering process.
 #' @inheritParams reporter_download_inputs
 #' @return `shiny::moduleServer`
+#' @details To access the default values for the `global_knitr` parameter,
+#'  use `getOption("teal.reporter.global_knitr")`.These defaults include:
```
m7pr, line 41:
```suggestion
#'  use `getOption("teal.reporter.global_knitr")`. These defaults include:
```

R/DownloadModule.R
```diff
@@ -33,9 +33,17 @@ download_report_button_ui <- function(id) {
...
+#' - `echo`: `TRUE`
```
m7pr, line 42:
```suggestion
#' - `echo = TRUE`
```

R/DownloadModule.R
```diff
@@ -33,9 +33,17 @@ download_report_button_ui <- function(id) {
...
+#' - `tidy.opts`: `list(width.cutoff = 60)`
```
m7pr, line 43:
```suggestion
#' - `tidy.opts = list(width.cutoff = 60)`
```

R/DownloadModule.R
```diff
@@ -33,9 +33,17 @@ download_report_button_ui <- function(id) {
...
+#' - `tidy`: This option is automatically set to `TRUE` if the `formatR` package is present in the namespace;
```
m7pr, line 44:
```suggestion
#' - `tidy = TRUE`  if `formatR` package is installed, `FALSE` otherwise
```

m7pr:
@kartikeyakirar please see my changes to one place where you provided those @details.
Can you make those `@details` a reusable template with `@template`? So that it is specified once, and not copy pasted in 4 places

R/zzz.R
```diff
@@ -0,0 +1,23 @@
+.onLoad <- function(libname, pkgname) {
+  # adapted from https://r-pkgs.org/code.html#sec-code-onLoad-onAttach
```
m7pr, line 2:
```suggestion

```

DESCRIPTION
```diff
@@ -31,6 +31,7 @@ Imports:
     zip (>= 1.1.0)
 Suggests:
     DT (>= 0.13),
+    formatR,
```
m7pr, line 34: hey @kartikeyakirar can you run verdepcheck flow to determine the minimal needed version of this package?
verdepcheck

kartikeyakirar:
> Can you make those @details a reusable template with @template?

@template is [superseded](https://roxygen2.r-lib.org/articles/reuse.html#superseded) instead I have defined the details in `simpleReporter.R` and reusing with inherit. thanks for suggestion.

m7pr:
@kartikeyakirar I wanted you to create roxygen template, not to inherit details.

Please subsitute `@inherit simple_reporter_srv details` with `#' @template roxlate-knitr-details`. Also put that in `R/SimpleReporter.R` instead of details.

Then create a file

`man-roxygen/roxlate-knitr-details.R` and put below in the content

```r
#' @details To access the default values for the `global_knitr` parameter,  use `getOption("teal.reporter.global_knitr")`. These defaults include:
#' - `echo = TRUE`
#' - `tidy.opts = list(width.cutoff = 60)`
#' - `tidy = TRUE`  if `formatR` package is installed, `FALSE` otherwise
```

m7pr:
@kartikeyakirar alrighty, they suggest
> Instead of @template and @templateVars write your own function and call it from inline R code.

Can you follow https://rmarkdown.rstudio.com/lesson-4.html ?
Looks like you can use `r global_details()` which would produce the content of details

m7pr (approved):
@kartikeyakirar looks good now. That was a long ride.
Would you be able to create a comment that proves that it works? The comparison of a behavior on `main` with this branch for the usage of all methods/functions that were changed?

kartikeyakirar:
> The comparison of a behavior on main with this branch for the usage of all methods/functions that were changed?
thanks for review @m7pr , I have updated description  for this.

m7pr:
coolio 👍

PR #232, opened 2023-10-20
define code_block function with quote not string

### LINKED ISSUE #221: Don't keep code as a string
Please change the below so that we don't keep executed code as a string:
https://github.com/insightsengineering/teal.reporter/blob/b82c2acb63cf2ef80e1009bdc030c8bafce91bfe/R/Renderer.R#L38-L55

If string is really needed then please convert code object into a string

chlebowa:
Closes #221

### DIFF
--- a/R/Renderer.R
+++ b/R/Renderer.R
@@ -37,30 +37,27 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
       }
 
       private$report_type <- get_yaml_field(yaml_header, "output")
-      format_code_block_function <- paste0(
-        c(
-          "code_block <- function (code_text) {",
-          "  df <- data.frame(code_text)",
-          "  ft <- flextable::flextable(df)",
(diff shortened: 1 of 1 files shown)


R/Renderer.R
```diff
@@ -37,30 +37,27 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
...
+        }
       )
 
       parsed_global_knitr <- sprintf(
         "\n```{r setup, include=FALSE}\nknitr::opts_chunk$set(%s)\n%s\n```\n",
         capture.output(dput(global_knitr)),
         if (identical(private$report_type, "powerpoint_presentation")) {
-          format_code_block_function
+          paste(deparse(format_code_block_function), collapse = "\n")
         } else {
```
pawelru, line 61: This is the only place in this method where we use `format_code_block_function`. Would it be good to move this obj definition here? That would help keep top function environment more clean.

pawelru (approved):
👍

PR #209, opened 2023-09-04
208 CRAN requirements for the first submission [skip vbump]

m7pr:
Closes #208 Closes #201
- satisfies DESCRIPTION requirements. 
- needed to remove `:::` from documentation - this required to 
    - update `teal.reporter:::ReportCard` to `teal.reporter::ReportCard` as this is exported object
    - update `teal.reporter:::Report` to `teal.reporter::Report` as this is exported object
    - delete `$append_content()` method examples as they used `teal.reporter:::NewpageBlock` which is unexported
    - delete manual pages for internal, unexported objects: `Archiver`, `ContentBlock`, `FileArchiver`, `FileBlock`, `JSONArchiver`, `NewpageBlock`, `PictureBlock`, `RcodeBlock`, `Renderer`
        - above can still be visible in `R/` directory in the source code of classes

PR #239, opened 2024-01-16
238 bring back internal examples

### LINKED ISSUE #238: Move internal examples from `dev/` folder into regular `R/*.R` files and use `getFromNamespace()` instead of `:::`
This file https://github.com/insightsengineering/teal.reporter/blob/main/dev/internal_examples.R was created out of multiple examples of internal functions that used `:::`. The usage of `:::` in CRAN is highly discouraged. Since we wanted to keep those examples for future developers, but not necessarily pack them in the package we moved them to `dev/` folder. `dev/` folder is listed in `.Rbuildignore` and is not included in R package during the build.

After an internal discussion about the same manner for `teal.code` package we decided to keep internal examples in their respective `R/*.R` files. However we decided to substitute `:::` with `getFromNamespace()` function.

m7pr:
Close #238 

Examples moved to `dev/` folder during https://github.com/insightsengineering/teal.reporter/pull/209 were brough back to their places. We substituted `:::` usage, that is discouraged by CRAN, with `getFromNamespace()` function


R/Archiver.R
```diff
@@ -6,6 +6,8 @@ Archiver <- R6::R6Class( # nolint: object_name_linter.
     #' @description Returns an `Archiver` object.
     #'
     #' @return an `Archiver` object
+    #' @examples
+    #' archiver <- getFromNamespace("Archiver", "teal.reporter")$new()
```
chlebowa, line 10: Please import the function/class and use imported.
```suggestion
    #' Archiver <- getFromNamespace("Archiver", "teal.reporter")
    #' Archiver$new()
```

chlebowa (changes requested):
I'm not sure if `$initialize` methods that take no arguments really require examples.

R/Archiver.R
m7pr: alrighty, that will take some time but is actually more visible for the end user on what's goin on : )
Will do one task first in here https://github.com/insightsengineering/teal.data/issues/238 and then get back to this one

m7pr: Changed this in 5d000820789beec4a2359841f148f8b9b3e0bb2b for other examples as well

R/Archiver.R
```diff
@@ -6,6 +6,9 @@ Archiver <- R6::R6Class( # nolint: object_name_linter.
     #' @description Returns an `Archiver` object.
     #'
     #' @return an `Archiver` object
+    #' @examples
+    #' archiver <- getFromNamespace("Archiver", "teal.reporter")
```
chlebowa, line 10: Maintain class name when importing.
```suggestion
    #' Archiver <- getFromNamespace("Archiver", "teal.reporter")
```

chlebowa (changes requested):

R/Archiver.R
m7pr: @chlebowa unsure if our lintr job will not update it afterwards to lowercase but will try

chlebowa: If linter complains, kill the linter. The examples must present the object in its natural form.

m7pr: ok it didnt : ) will change where needed

m7pr: Hey, incorporated this update. I think it's way easier to track if we use class names. Thanks

chlebowa (approved):

PR #252, opened 2024-02-21
187 color scheme for disabled button

### LINKED ISSUE #187: Color scheme for disabled button
> Felt like we can do better when the button is disabled (blue and grey just not pleasing to the eye).
> 
> image
> 
> image
> 

_Originally posted by @donyunardi in https://github.com/insightsengineering/teal.reporter/issues/186#issuecomment-1550676354_

chlebowa:
Closes #187 

Updated background color of two buttons (anchor elements) identified by _partial ids_ in disabled state.

The background color is `darkgrey` nad text color is `white`.

image
image

### DIFF
--- a/inst/css/Previewer.css
+++ b/inst/css/Previewer.css
@@ -14,10 +14,10 @@ a.disabled {
   text-decoration: none;
 }
 
-[id$=download_data_prev].disabled {
-  background-color: darkgrey;
-}
-[id$=download_data].disabled {
+a[id$=download_data].disabled,
+a[id$=download_data_prev].disabled {
+  border: 0;
+  color: white;
   background-color: darkgrey;
 }
 
--- a/inst/css/Previewer.css
+++ b/inst/css/Previewer.css
@@ -14,6 +14,13 @@ a.disabled {
   text-decoration: none;
 }
 
+[id$=download_data_prev].disabled {
+  background-color: darkgrey;
+}
+[id$=download_data].disabled {
+  background-color: darkgrey;
+}
+
 /* icons in the previewer */
 .icon_previewer {
   float:right;


Polkas:
Please be careful and check it out on all 3 bootstrap versions.
The issue is connected only with bootstrap version 3. 
However, the code update will influence all of them (3, 4 and 5).
We should be careful with the id selector as we can have an app with a few such modules, and then we look for a class rule.

Polkas:
the download buttons are a tags in Shiny, and the solution is probably to edit the linked rule from the same file.

```r
/* Disable any anchor with disabled class */
a.disabled {
  color: grey;
  cursor: not-allowed;
  pointer-events: none;
  text-decoration: none;
}
```

chlebowa:
> check it out on all 3 bootstrap versions

Good point :+1: 

> A class rule seems more natural than an id suffix-based selector.

This was intentional. I didn't want to affect other `a` elements because, frankly, I wasn't sure how many others there are. This solution affects these two download buttons only.

Polkas:
The `a.disabled` rule is already in the css file and narrows the scope only to disabled a tags. So please consider my proposition.

chlebowa:
Another good point :slightly_smiling_face:

chlebowa:
Well, styling by class only seems to work with BS 3 :shrug: 
Styling by id works in BS 3, 4, and 5.

Polkas:
I assumed we wanted to fix Bootstrap 3 only, and omitting a rule for versions 4 and 5 seemed ok to me.
If we want to apply it to all of them, we must use ! an `!important` solution.

```css
/* Disable any anchor with disabled class */
a.disabled {
  color: white !important;
  cursor: not-allowed;
  pointer-events: none;
  text-decoration: none;
}
```

chlebowa:
I am wary of using `!important` on a tag. There is a risk of the style spilling out to other elements (`a` tags in other modules). I think it is safer to narrow down the specification by id.

How about `a[id$=download_data].disabled`?

Polkas:
Looks nice now:)
Bootstrap 5
Screenshot 2024-02-22 at 17 14 19
Bootstrap 4
Screenshot 2024-02-22 at 17 13 33
Bootstrap 3
Screenshot 2024-02-22 at 17 11 35

chlebowa:
Awesome, thanks for checking :+1: :slightly_smiling_face:

gogonzo (approved):

PR #264, opened 2024-03-15
set to release

gogonzo:
Fixes #256 

Issue reported on [CRAN](https://www.stats.ox.ac.uk/pub/bdr/M1mac/teal.reporter.out) is reproducible also on our `insightsengineering` image ([see here](https://github.com/insightsengineering/teal.reporter/pull/263)).
Current commit doesn't fail on the same image.

PR #266, opened 2024-03-19
[skip actions] [skip vbump] Revert 2 commits

donyunardi:
This reverts commit ef1f79f4ba83e93db113610e557ab8f2a3f09075 and 384f9414.

We need to revert these commits so that we can merge #264, ensuring that we have the same content in both CRAN and GitHub tags.

We will re-revert the commits after the release.


vedhav (approved):
👌🏽

PR #268, opened 2024-03-20
898 save app state version 3

chlebowa:
Companion to https://github.com/insightsengineering/teal/pull/1011
Adds `shiny::restoreInput` calls to the `value`/`choices` arguments in `update*Input` calls.

### DIFF
--- a/R/Previewer.R
+++ b/R/Previewer.R
@@ -77,10 +77,15 @@ reporter_previewer_srv <- function(id,
   )
   checkmate::assert_true(rmd_yaml_args[["output"]] %in% rmd_output)
 
-  shiny::moduleServer(
-    id,
-    function(input, output, session) {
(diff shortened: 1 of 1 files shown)


averissimo:
Simple change `expect_snapshot(app$ui)` to `expect_snapshot(app$ui())` as `ui` element is now a function _(due to bookmarking)_

No other package seems to use this type of snapshot, so the issue was unique to `{t.m.helios}`

gogonzo (approved):
change could be independent of teal. Works well 👍

Polkas:
I got some unexpected add card reactivity:(

Uploading Screen Recording 2024-03-26 at 12.13.27 AM.mov…

chlebowa:
This is expected and will be dealt with shortly.

PR #273, opened 2024-05-16
Bumps `{rmarkdown}` minimal version

averissimo:
# Pull Request

Part of https://github.com/insightsengineering/coredev-tasks/issues/546

Necessary bump to overcome a lack of binary on ppm snapshots that causes an error on minimum strategies for `scheduled` workflows.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,7 @@ Imports:
     R6,
     rlistings (>= 0.2.8),
     rmarkdown (>= 2.23),
-    rtables (>= 0.5.1),
+    rtables (>= 0.6.6),
     shiny (>= 1.6.0),
     shinybusy (>= 0.3.2),
     shinyWidgets (>= 0.5.1),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,7 @@ Suggests:
     DT (>= 0.13),
     formatR (>= 1.5),
     formatters,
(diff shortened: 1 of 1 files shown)


averissimo:

averissimo:
`{verdepcheck}` strategies pass: https://github.com/insightsengineering/teal.reporter/actions/runs/9398895960/job/25885216394

Exception is `max`, but that is due to `Suggests` dependency `rtables` latest from Github requires a version of `formatters` that is not yet released.

This is ok to merge!

DESCRIPTION
```diff
@@ -53,7 +53,7 @@ VignetteBuilder:
 RdMacros:
     lifecycle
 Config/Needs/verdepcheck: rstudio/bslib, mllg/checkmate,
-    davidgohel/flextable, cran/grid, rstudio/htmltools, yihui/knitr,
```
kartikeyakirar, line 56: grid is removed from cran right? this package is in import

averissimo: This is a base package. The `cran/grid` is 22 years old and it will break this

averissimo: I mena, it will break this if we add the github mirror as it's not the "real" code we want to test

kartikeyakirar: oh okay got it.

kartikeyakirar (approved):

PR #285, opened 2024-08-27
Fix failed pipeline

### LINKED ISSUE #284: Fix failed pipeline
https://github.com/pawelru/ie_status?tab=readme-ov-file

donyunardi:
Fixes #284 

Small note, I ended up using `testthat::skip_if_not_installed("ggplot2")` at the beginning of several unit test files (e.g., `test-DownloadReportModule.R`, `test-LoadReporterModule.R`, etc.) because they depend on `ggplot2` at the start of the test. 

I don't think this update will affect our code coverage since code coverage is done using a different workflow (`Check`) which I believe will install all dependencies, including `ggplot2`.

### DIFF
--- a/R/Reporter.R
+++ b/R/Reporter.R
@@ -25,7 +25,7 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
     #'
     #' @param cards (`ReportCard`) or a list of such objects
     #' @return `self`, invisibly.
-    #' @examples
+    #' @examplesIf require("ggplot2")
     #' library(ggplot2)
     #' library(rtables)
     #'
@@ -57,7 +57,7 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
     #' @description Retrieves all `ReportCard` objects contained in the `Reporter`.
     #'
     #' @return A (`list`) of [`ReportCard`] objects.
-    #' @examples
+    #' @examplesIf require("ggplot2")
     #' library(ggplot2)
     #' library(rtables)
     #'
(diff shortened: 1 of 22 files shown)


donyunardi:

donyunardi:
Scheduled workflow passed:
https://github.com/insightsengineering/teal.reporter/actions/runs/10586868417

m7pr:
`ggplot2` is in Suggests so such a workaround with `skip_if_not_installed()` is recommended.

Few points:
- I don't understand the addition of `rmarkdown` to `VignetteBuilder` honestly and for now I don't think it should be there.
- @donyunardi should we rewrite few tests so that they use regular base plots, so that we don't skip them?

m7pr:
Or is `rmarkdown` here just to satisfy the `noSuggest` check where `rmarkdown` was not found and when it's in `VignetteBuilder` then it's found? If this is the only way to do that, then this is fine

R/Reporter.R
```diff
@@ -57,7 +57,7 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
     #' @description Retrieves all `ReportCard` objects contained in the `Reporter`.
     #'
     #' @return A (`list`) of [`ReportCard`] objects.
-    #' @examples
+    #' @examplesIf require("ggplot2")
```
m7pr, line 60: @donyunardi in other packages (like `teal.slice`) we have `requireNamespace` in @examplesIf

```suggestion
    #' @examplesIf requireNamespace("ggplot2")
```

donyunardi:
> Or is rmarkdown here just to satisfy the noSuggest check where rmarkdown was not found and when it's in VignetteBuilder then it's found? If this is the only way to do that, then this is fine

Yes, @pawelru did the research [here](https://github.com/insightsengineering/coredev-tasks/issues/577#issuecomment-2295842071)

> @donyunardi should we rewrite few tests so that they use regular base plots, so that we don't skip them?

That was my first impression too, but then we're only skipping this for workflow like noSuggest. Since we have other workflow with all dependencies installed, these tests are still get executed. 

If we want to upgrade these test cases to use regular base plots, I support this but I'd prefer to handle it outside of this issue since that would require more attention.

R/Reporter.R
m7pr: Alrighty, so I think we should update other packages where it got `requireNamespace`

m7pr (approved):

PR #289, opened 2024-11-18
Add `rtables.officer` as a dependency

vedhav:
Changes:
1. Updates the `rtables` version
2. Adds `rtables.officer` as a new dependency

Note that this blocks CRAN release of `teal.reporter` untill the `rtables.officer` and `rtables` is released.

### DIFF
--- a/staged_dependencies.yaml
+++ b/staged_dependencies.yaml
@@ -25,6 +25,9 @@ upstream_repos:
   insightsengineering/rtables:
     repo: insightsengineering/rtables
     host: https://github.com
+  insightsengineering/rtables.officer:
+    repo: insightsengineering/rtables.officer
+    host: https://github.com
   insightsengineering/rlistings:
     repo: insightsengineering/rlistings
     host: https://github.com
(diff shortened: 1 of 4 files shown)


staged_dependencies.yaml
```diff
@@ -25,6 +25,9 @@ upstream_repos:
   insightsengineering/rtables:
     repo: insightsengineering/rtables
     host: https://github.com
+  insightsengineering/rtables.officer:
+    repo: insightsengineering/rtables.officer
+    host: https://github.com
```
shajoezhu, line 30: i thought you guys have stopped using staged.dependnecies?

shajoezhu (approved):
thanks a lot @vedhav for the changes, I think it all looks good. i was wondering if the new changes in rtables, is blocking teal package development? If not, can we work together with the release plan. Thanks!

staged_dependencies.yaml
vedhav: That's correct, we're in the middle of migrating ([so far only teal is migrated](https://github.com/insightsengineering/nestdevs-tasks/issues/65)), But, we're working on other tasks so it is on hold until we get to it.

vedhav:
Thanks @shajoezhu we do not plan to release the `teal.reporter` this increment so we can follow the `rtables` release plan.

PR #290, opened 2024-11-26
Remove vignette note

llrs-roche:
# Pull Request

[CI checks are failing](https://github.com/r-universe/pharmaverse/actions/runs/11851042359/job/33026981747) due to the vignettes not being created.

There is a note about using a Rmd as childe file in the vignettes. This file is recognized as a vignette.
We can a) move it to a different folder like what it is done for articles, something like `vignettes/child`? or b) move the code chunk into the vignette.
As it was only used once, I moved the code chunk into the file directly. 
This made this error disappear. 

There were errors due to some code moving from rtables to rtables.office. I think this should be taken care now that rtables.office is [fixed](https://github.com/insightsengineering/rtables.officer/pull/21) and building well.


llrs-roche:

PR #298, opened 2025-01-14
upgrade formatters and rlistings

m7pr:
There is also one more tests failing on dependency tests due to formatters+rlistings combination

```r
 ── Error ('test-utils.R:25'): to_flextable: supported class `listing_df` ───────
  Error in `FUN(X[[i]], ...)`: argument "fontspec" is missing, with no default
  Backtrace:
       ▆
    1. └─teal.reporter:::to_flextable(lsting) at test-utils.R:25:3
    2.   ├─rlistings::matrix_form(content)
    3.   └─rlistings::matrix_form(content)
    4.     └─rlistings (local) .local(obj, indent_rownames, expand_newlines)
    5.       ├─formatters::MatrixPrintForm(...)
    6.       │ └─base::structure(...)
    7.       ├─formatters::make_row_df(obj)
    8.       └─rlistings::make_row_df(obj)
    9.         └─rlistings (local) .local(...)
   10.           ├─rlistings:::vec_nlines(tt[[col]], max_width = colwidths[col])
   11.           └─rlistings:::vec_nlines(tt[[col]], max_width = colwidths[col])
   12.             ├─base::unlist(lapply(format_colvector(colvec = vec), nlines, max_width = max_width))
   13.             └─base::lapply(format_colvector(colvec = vec), nlines, max_width = max_width)
   14.               ├─formatters (local) FUN(X[[i]], ...)
   15.               └─formatters (local) FUN(X[[i]], ...)
   16.                 └─base::vapply(...)
   17.                   └─formatters (local) FUN(X[[i]], ...)
   18.                     └─formatters::wrap_txt(xi, max_width, fontspec = fontspec)
   19.                       └─formatters::open_font_dev(fontspec)
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,7 @@ Imports:
 Suggests:
     DT (>= 0.13),
     formatR (>= 1.5),
-    formatters,
+    formatters (>= 0.5.10),
     ggplot2 (>= 3.4.3),
     lattice (>= 0.18-4),
     png,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,7 +30,7 @@ Imports:
     knitr (>= 1.42),
     lifecycle (>= 0.2.0),
     R6,
-    rlistings (>= 0.2.8),
+    rlistings (>= 0.2.10),
     rmarkdown (>= 2.23),
     rtables (>= 0.6.10.9004),
     rtables.officer (>= 0.0.1.9005),


llrs-roche (approved):
Do we really need 0.5.10? It seems formatters 0.5.8 already introduced the `fontspec` argument on `wrap_txt` that is needed according to the tests: https://diffify.com/R/formatters/0.5.5/0.5.8

m7pr:
dependency test action https://github.com/insightsengineering/teal.reporter/actions/runs/12768063698

m7pr:
New job https://github.com/insightsengineering/teal.reporter/actions/runs/12769114391

llrs-roche:
I see that you added rlisting version too.
But the actions are still failing with:

```
Error in FUN(X[[i]], ...) : subscript out of bounds
Calls: fun ... new_release_deps_installation_proposal -> map_key_character -> vapply
Execution halted
Ran 1/1 deferred expressions
```

Triggering a new run

llrs-roche (changes requested):
The new actions still fail with at `⠴ Resolving release version of withr [25/25]   ETA: 0s`

m7pr:
Yeah but it fails for `release` strategy and we accept this strategy might fail. This test the package only with released dependencies, it does not use development dependencies. Sometimes package only works with the development dependencies and will no longer work with only released packages (that's why we need to keep releasing more often).

I was trying to fix the other strategy (`min_isolated`), which this PR actually fixed.

Would you mind accepting?

llrs-roche (approved):
Oh, I didn't know that sometimes the release branch is allowed to break. All good then.

m7pr:
@llrs-roche so think about it like that

we have released packages:

teal.reporter ver 0.3.0 and it's dependency rtables 0.2.0

When rtables gets an improvement to the version 0.2.0.9001 but does not yet gets released, you sometimes need to satisfy new features in teal.reporter and you bump the dependency version to rtables (>= 0.2.0.9001). So now teal.reporter is dependent on the development version of the rtables. However release strategy only checks released packages, and the last released version is 0.2.0 for which teal.reporter breaks. So untill all dependencies are not released, the release strategy will keep breaking.

PR #299, opened 2025-01-21
Match code and documentation

llrs-roche:
While checking for other issues I noticed in the examples some mismatch between the text and the result. 
I also observed some differences between the options on the code and the documented options. 

While checking the package I saw:

```
─  checking examples ... [22s] OK (23.2s)
   Examples with CPU (user + system) or elapsed time > 5s
            user system elapsed
   Reporter 2.61   0.14    7.09
   Renderer 2.30   0.30    7.74
```

The problem might come from each method creating a new card and reporter, see https://github.com/r-lib/roxygen2/issues/1521 for the underlying issue and linked issues here.


llrs-roche:
Thanks Vedha! I updated `rtables.officer` and `rtables` versions

m7pr (approved):

Issue #280, opened 2024-06-27
Improve the Reproducibility Mechanisms and Code Display in Report Output


donyunardi:
## Improve Reproducibility Mechanism
When a user utilizes the `teal.reporter` feature in their teal app and clicks the `Download Report` button from the Reporter previewer module, they will receive:
* The output file (PDF, HTML, PPT, Word)
* PNG or RDS file
* Report.json file
* Rmd file

The Rmd file that the user receives will not generate new output (plot or table) if they change the R code section in the Rmd because:
1. The code is not re-evaluated when the markdown is knitted, as the R code chunk option is set to `eval=FALSE`.
2. The image is embedded in the Rmd file using `knitr::include_graphics` or `readRDS`.

This means the Rmd file is static and the plots or tables are not reproducible. To make the Rmd reproducible, new output should be generated when it's knitted. Outputs should not be embedded (as in step 2) in the Rmd file.

## Improve Code Display
We also want to improve how the code is displayed in the output report (and possibly in Rmd) because it is currently shown as one long line of code, regardless of the programming steps (e.g., data processing, filtering, plotting, etc.).

Users will have more clarity and readability if we can provide sections for the code in the report.

## Definition of Done
- A reproducible Rmd file means that knitting the Rmd will create new outputs.
- Better code readability in the report, allowing users to distinguish between sections of code for data processing, filtering, etc.

Issue #248, opened 2024-02-16
[Feature Request]: Skip examples / tests / vignettes on Suggest packages


averissimo:
### Feature description

Needs to:

- Convert examples tags to  `#' @examplesIf requireNamespace("pkgname")`
  - For `ggplot2` & `rtables`
  - ⚠️ `@examplesIf` isn't supported in R6 methods documentation
    - (see https://github.com/r-lib/roxygen2/issues/1521)
- Add an `eval` property to chunks with same logic and using the `quietly = TRUE` argument
  - Can be easily done with `knitr::opts_chunk$set`
- Skip tests that use `ggplot2`
  - Has some effort as tests are not self-contained (see `test-Archiver.R` and diff below)

<details>
<summary>Prototype diff patch to improve on `test-Archiver.R`</summary>

```diff
diff --git a/tests/testthat/test-Archiver.R b/tests/testthat/test-Archiver.R
index aeee449..a791b79 100644
--- a/tests/testthat/test-Archiver.R
+++ b/tests/testthat/test-Archiver.R
@@ -1,23 +1,39 @@
-card1 <- ReportCard$new()
 
-card1$append_text("Header 2 text", "header2")
-card1$append_text("A paragraph of default text", "header2")
-card1$append_plot(
-  ggplot2::ggplot(iris, ggplot2::aes(x = Petal.Length)) +
-    ggplot2::geom_histogram()
-)
+# Helper to generate a test reporter on demand instead of common shared object
+build_test_reporter <- function() {
+  card1 <- ReportCard$new()
 
-card2 <- ReportCard$new()
+  card1$append_text("Header 2 text", "header2")
+  card1$append_text("A paragraph of default text", "header2")
+  card1$append_plot(
+    ggplot2::ggplot(iris, ggplot2::aes(x = Petal.Length)) +
+      ggplot2::geom_histogram()
+  )
+
+  card2 <- ReportCard$new()
 
-card2$append_text("Header 2 text", "header2")
-card2$append_text("A paragraph of default text", "header2")
-lyt <- rtables::analyze(rtables::split_rows_by(rtables::basic_table(), "Day"), "Ozone", afun = mean)
-table_res2 <- rtables::build_table(lyt, airquality)
-card2$append_table(table_res2)
-card2$append_table(iris)
+  card2$append_text("Header 2 text", "header2")
+  card2$append_text("A paragraph of default text", "header2")
[...]

PR #300, opened 2025-01-22
[skip vbump] upversion to 0.4.0

donyunardi:
Related to #296 

* Update NEWS 
* Update DESCRIPTION and upversion rtables.office and rtables to released version


m7pr (approved):

donyunardi:
I removed `Remotes` entry for `rtables.officer` because it's not available in CRAN.

donyunardi:
Submitted to CRAN. Awaiting Feedback.

donyunardi:
Got automated feedback not passing the R-devel tests:
https://win-builder.r-project.org/incoming_pretest/teal.reporter_0.4.0_20250123_012612/Debian/00check.log

```
* checking examples ... [19s/19s] NOTE
Examples with CPU (user + system) or elapsed time > 5s
          user system elapsed
Reporter 6.788  0.032   6.820
Renderer 5.886  0.172   6.097
```

That's odd, this passed previously.

donyunardi:
I see `iris` is being used on the test cases of Renderer and Reporter. One way that I could think of is to make a smaller subset of `iris` or use other dataset like `cars`.

@insightsengineering/nest-core-dev what do you think?

llrs-roche:
I found this same problem on #299 , I explored it a bit and I think it is a problem with roxygen @examplesIf, see also #248. I don't think changing datasets will reduce computing time. We might be able to reduce (more)time if instead of having each method create a new card and reporter we reused one. It might be harder to follow individual methods examples but it will probably make the examples faster.

donyunardi:
It turns out, we have the same message in our r-hub action as well in the main branch (Debian == gcc12):
https://github.com/insightsengineering/teal.reporter/actions/runs/12850103218/job/35829405265#step:8:119

After investigation, I decided to just remove the second `append_table()` method for iris dataset. 

The pipeline is now completed without NOTE:
https://github.com/insightsengineering/teal.reporter/actions/runs/12940202534/job/36093952669#step:8:118

m7pr (approved):

pawelru:
It would be the best to use profiler tools to identify performance bottlenecks. I think that you are guessing here.

```
pkgload::run_example("./man/Reporter.Rd") |> profvis::profvis()
```

Currently the main performance bottleneck is converting rtables to flextable (which is btw happening outside of this package). Hance I would rather remove / shorten rtables related example.

image

pawelru:
Sorry I was too fast and made mistake

rtables to flextable is responsible for this:
image

Whereas data.frame to flextable is responsible for this:
image

In this case - reducing examples with data.frames would actually help. I don't know if the size matter though...

BTW: why do force conversion of data.frame to flextable? It looks like an expensive operation.

vedhav:
> why do force conversion of data.frame to flextable? It looks like an expensive operation.

I'd have to check, but it could be because of the output render formats. flextable renders in all the formats easily, I'm not sure about rtables.

gogonzo:
> BTW: why do force conversion of data.frame to flextable? It looks like an expensive operation.

data.frame needs to be converted to display it in the document. There are several design pitfalls, the most serious:
- too early conversion during *Block initialization (via $set_content). It should rather be converted when object is rendered (in Renderer or previewer). Package needs refactor and some discussions took place in a hotel lobby during useR! last year ;] (with @chlebowa @averissimo )

> I'd have to check, but it could be because of the output render formats. flextable renders in all the formats easily, I'm not sure about rtables.

`flextable` is used to "easily" convert object to Rmd and HTML (see `Renderer$renderRmd`, `block_to_html()`)

pawelru:
> too early conversion during *Block initialization (via $set_content). It should rather be converted when object is rendered (in Renderer or previewer).

Yeah that's exactly it. Probably not worth to postpone the release because of that but I think it's important.

Also posting this just to show you how to handle "examples takes too long" type of problems.

gogonzo:
> Yeah that's exactly it. Probably not worth to postpone the release because of that but I think it's important.

It is not justv about TableBlock. Plots for example are saved into file during PictureBlock initialization. I'm not a fan to do these changes now as we already know that the whole package needs to be revisited.

donyunardi:
Thanks @pawelru! You advise will come in handy moving forward if we run into this type of issue again when submitting to CRAN.

Also, as you mentioned, refactoring the examples is not worth to postpone the release, especially we're thinking about refactoring teal.reporter with [this](https://github.com/insightsengineering/teal.reporter/issues/280) issue

I will move forward with the release.

donyunardi:
Submitted. Awaiting CRAN feedback.

donyunardi:
made it to CRAN
image

PR #307, opened 2025-02-24
Redesign of `teal.reporter` Cards

m7pr:
### Companion to  

- [teal.modules.general/report_redesign_poc@main](https://github.com/insightsengineering/teal.modules.general/tree/report_redesign_poc%40main)  
- [teal/redesign@main](https://github.com/insightsengineering/teal/tree/redesign%40main)
- [teal.code/redesign@main](https://github.com/insightsengineering/teal.code/tree/redesign%40main)

Closes:

- https://github.com/insightsengineering/nestdevs-tasks/issues/106

<details><summary>Code</summary>  

```r

devtools::load_all('../teal.reporter')
devtools::load_all('../teal.widgets')
devtools::load_all('../teal.code')
devtools::load_all('../teal')
devtools::load_all('.')

# general data example
data <- teal_data()
data <- within(data, {
  require(nestcolor)
  CO2 <- CO2
  USArrests <- USArrests
})

app <- init(
  data = data,
  modules = modules(
    tm_a_regression(
      label = "Regression",
      response = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = "uptake",
          selected = "uptake",
          multiple = FALSE,
          fixed = TRUE
        )
      ),
      regressor = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variables:",
          choices = variable_choices(data[["CO2"]], c("conc", "Treatment")),
          selected = "conc",
          multiple = TRUE,
          fixed = FALSE
        )
      )
    ),
    tm_a_pca(
      "PCA",
      dat = data_extract_spec(
        dataname = "USArrests",
        select = select_spec(
          choices = variable_choices(
            data = data[["USArrests"]], c("Murder", "Assault", "UrbanPop", "Rape")
          ),
          selected = c("Murder", "Assault"),
          multiple = TRUE
        ),
        filter = NULL
      )
    ),
    tm_g_scatterplot(
      label = "Scatterplot Choices",
      x = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(data[["CO2"]], c("conc", "uptake")),
          selected = "conc",
          multiple = FALSE,
          fixed = FALSE
        )
      ),
      y = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(data[["CO2"]], c("conc", "uptake")),
          selected = "uptake",
          multiple = FALSE,
          fixed = FALSE
        )
      ),
      color_by = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(
            data[["CO2"]],
            c("Plant", "Type", "Treatment", "conc", "uptake")
          ),
          selected = NULL,
          multiple = FALSE,
          fixed = FALSE
        )
      ),
      size_by = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(data[["CO2"]], c("conc", "uptake")),
          selected = "uptake",
          multiple = FALSE,
          fixed = FALSE
        )
      ),
      row_facet = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(data[["CO2"]], c("Plant", "Type", "Treatment")),
          selected = NULL,
          multiple = FALSE,
          fixed = FALSE
        )
      ),
      col_facet = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(data[["CO2"]], c("Plant", "Type", "Treatment")),
          selected = NULL,
          multiple = FALSE,
          fixed = FALSE
        )
      )
    ),
    tm_g_bivariate(
      x = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(data[["CO2"]]),
          selected = "conc",
          fixed = FALSE
        )
      ),
      y = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(data[["CO2"]]),
          selected = "uptake",
          multiple = FALSE,
          fixed = FALSE
        )
      ),
      row_facet = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(data[["CO2"]]),
          selected = "Type",
          fixed = FALSE
        )
      ),
      col_facet = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = variable_choices(data[["CO2"]]),
          selected = "Treatment",
          fixed = FALSE
        )
      )
    )
  )
)
if (interactive()) {
  shinyApp(app$ui, app$server, enableBookmarking = "server")
}
```

</details>  

---

## Cards  

This PR introduces a new way of creating and using `teal.reporter` Cards.  

`teal.reporter` Cards store snapshots of `teal` module outputs - referred to as `ReportCards` (or simply "Cards").  

### Current usage of Cards:  
- **Creating Cards** – Triggered by the "Add Card" button.  
- **Previewing Cards in HTML** – Rendered in the Reporter Preview(er) tab.  
- **Storing and Restoring Cards** – Managed via Report Download and Report Upload, involving the Renderer and Archiver.  
- **Rendering Cards to (R)Markdown** – Used when downloading a report.  
- **Editing Cards** – It is possible within the Reporter Preview(er) tab to edit Report Cards.  

---

## Introducing `ReportDocument`  

This PR introduces `ReportDocument`, an alternative to `ReportCard`, with the following advantages:  

- **Simplified creation** – `ReportDocument` is implemented as an S3 list, making it easier to create.  
- **Reduced storage size** – Unlike `ReportCard` (which is an R6 class and stores methods and parent classes), `ReportDocument` uses simple S3 storage, consuming less disk space.  
- **Flexible modifications** – Supports reordering, subsetting, and editing:  
  - **At the module level** – App developers can modify reports.  
  - **In the Reporter Preview(er) tab** – App users can edit reports.  

---

### Key Components  
- `report_document()` – Creates a new S3 class called `ReportDocument`, replacing `ReportCard`.  
- `ReportCard` – Previously contained various parts of the report, such as tables, plots, text, code, and HTML (`ContentBlock`s).  
- `ReportDocument` – A lightweight S3 list that can store characters, data frames, plots, and code blocks.  

### Process Overview  
1. **Adding a ReportDocument**  
   - When the `+` button is clicked in `teal/module_nested_tabs.R` via `teal.reporter::add_card_button_ui`, a `ReportDocument` is added to `Reporter` `$cards` using `teal::add_document_button_srv`.  

2. **Previewing a Report**  
   - Opening the "Report Preview(er)" tab triggers `reporter_previewer_srv`, which converts `ReportDocument` elements into HTML via `block_to_html`.

3. **Downloading a Report**  
   - Clicking "Download Report" in the previewer runs `report_render_and_compress` on `Reporter`, using `Renderer` to generate an `.Rmd` file.  
   - A `.zip` file is created, containing the `.Rmd` report and a JSON file for restoring the report.  

4. **Loading a Report**  
   - Clicking "Load Report" in the previewer clears the current report and restores it from the JSON file inside the `.zip`.  

5. **Editing a ReportDocument**
   - Clicking the "Edit" button on a `ReportDocument` Card allows users to modify text elements.  
   - User can specify custom edit functionality for specific classes

---

# `ReportDocument` usage

We expose `c.ReportDocument`, `[.ReportDocument` that are used inside `append`, so that user can combine, subset and append the object. However S3 methods are dispatched by the class of the first element, so `ReportDocument` needs to be the first element of the call of those functions. That's why we created `edit_document_content` that can modify and append and that keeps the class of `ReportDocument`.

``` r
library(teal.reporter)

######### Playing with ReportDocument
report <- report_document()
class(report)
#> [1] "ReportDocument"
report <- c(report, list("## Headline"), list("## Table"), list(summary(iris)))
report <- report[1:2]
report <- append(report, c(list("## Table 2"), list(summary(mtcars))), after = 1)
class(report)
#> [1] "ReportDocument"

######### Designed function for editing
report <- report_document(1, 2, 'c')
new_report <- edit_report_document(report, modify = c(3, 1), append = 'd')
new_report
#> [[1]]
#> [1] "c"
#> 
#> [[2]]
#> [1] 1
#> 
#> [[3]]
#> [1] "d"
#> 
#> attr(,"class")
#> [1] "ReportDocument"
class(new_report)
#> [1] "ReportDocument"
```

<sup>Created on 2025-03-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

# `ReportDocument` usage in `teal` modules

No need to know about 
- R6 structures inside `teal.reporter`
- `teal::report_card_template`
- names of `ReportCard` methods (`set_name`, `append_text`, `append_plot`, `append_src`) 

https://github.com/insightsengineering/teal.modules.general/pull/862/files?diff=split&w=0
image

# Comparison 

## Backend code to Render a document

```r
# SAME START
devtools::load_all(".")
devtools::load_all("../teal")
library(ggplot2)

reporter <- Reporter$new()

# OLD WAY
old_report_card <- ReportCard$new()
old_report_text_block <- TextBlock$new()
old_report_card$append_content(old_report_text_block)
old_report_card$set_name('NameOld')
reporter$append_cards(list(old_report_card))

# NEW WAY
new_report_card <- report_document('## Plot', ggplot2::qplot(1:5, 1:5))
new_report_card_markdown <- teal:::to_markdown(new_report_card) # this happens in teal::add_document_button_srv
names(new_report_card_markdown) <- 'NewName'                    # this happens in teal::add_document_button_srv
# above 2 steps will be moved to the rendering part
reporter$append_cards(new_report_card_markdown)

# SAME END

reporter$get_cards()
reporter$get_blocks()

input_list <- 
  list(
    author = "NEST", 
    title = "Report", 
    date = structure(20146, class = "Date"), 
    output = "html_document", 
    toc = FALSE
  )
global_knitr <- 
  list(
    echo = TRUE, 
    tidy.opts = list(width.cutoff = 60), 
    tidy = TRUE
  )
path <- 'render_output.zip'
report_render_and_compress(reporter, input_list, global_knitr, path)

```

## UI differences

- Module specific Add Card button
- Global Load Report button, instead of module specific
- Global Reset Report button, instead of module specific
- Global Download Report button, instead of module specific
- Removed download report encoding panel from Report previewer

---

## Files to be remove after deprecation

- `R/ReportCard.R`
- `R/ContentBlock.R`
- `R/FileBlock.R`
- `R/HTMLBlock.R`
- `R/NewpageBlock.R`
- `R/PictureBlock.R`
- `R/RcodeBlock.R`
- `R/TableBlock.R`
- `R/TextBlock.R`
- `R/SimpleReporter.R`
- `tests/testthat/test-ContentBlock.R`
- `tests/testthat/test-FileBlock.R`
- `tests/testthat/test-HTMLBlock.R`
- `tests/testthat/test-NewpageBlock.R`
- `tests/testthat/test-PictureBlock.R`
- `tests/testthat/test-RcodeBlock.R`
- `tests/testthat/test-TableBlock.R`
- `tests/testthat/test-TextBlock.R`

---

### DIFF
--- a/tests/testthat/helper-Reporter.R
+++ b/tests/testthat/helper-Reporter.R
@@ -29,22 +29,27 @@ test_card2.ReportCard <- local({ # nolint: object_name.
   }
 })
 
-test_card1 <- function() {
+test_card1 <- function(title = NULL) {
   withr::with_environment(emptyenv(), plot <- ggplot2::ggplot(iris, ggplot2::aes(x = Petal.Length)) +
     ggplot2::geom_histogram(binwidth = 0.2))
-  report_document("## Header 2 text", "A paragraph of default text", plot)
+  new_card <- report_document("## Header 2 text", "A paragraph of default text", plot)
+  if (!is.null(title)) metadata(new_card, "title") <- title
+  new_card
 }
 
 test_card2 <- local({
-  fun <- function() {
+  fun <- function(title = NULL) {
     lyt <- rtables::analyze(rtables::split_rows_by(rtables::basic_table(), "Day"), "Ozone", afun = mean)
     table_res2 <- rtables::build_table(lyt, within(airquality, Day <- factor(Day))) # nolint: object_name.
-    report_document("## Header 2 text", "A paragraph of default text", table_res2, iris)
+    new_card <- report_document("## Header 2 text", "A paragraph of default text", table_res2, iris)
+    if (!is.null(title)) metadata(new_card, "title") <- title
+    new_card
   }
-  cache <- NULL
-  function() {
-    if (is.null(cache)) cache <<- fun()
-    cache
+  cache <- list()
+  function(title = NULL) {
+    title_ix <- title %||% "no_title"
+    if (is.null(cache[[title_ix]])) cache[[title_ix]] <<- fun(title)
+    cache[[title_ix]]
   }
 })
 
--- a/tests/testthat/test-Reporter.R
+++ b/tests/testthat/test-Reporter.R
@@ -58,15 +58,27 @@ testthat::test_that("get_cards returns the same cards which was added to reporte
 })
 
 testthat::test_that("get_blocks returns the same blocks which was added to reporter, sep = NULL", {
-  reporter <- test_reporter(card1 <- test_card1(), card2 <- test_card2())
-  testthat::expect_identical(reporter$get_blocks(sep = NULL), append(unclass(card1), unclass(card2)))
+  reporter <- test_reporter(card1 <- test_card1("A title"), card2 <- test_card2("Another title"))
+  testthat::expect_identical(
+    reporter$get_blocks(sep = NULL),
+    append(
+      c(sprintf("# %s", metadata(card1, "title")), card1),
+      c(sprintf("# %s", metadata(card2, "title")), card2)
+    )
+  )
 })
 
 testthat::test_that("get_blocks by default adds NewpageBlock$new() between cards", {
-  reporter <- test_reporter(card1 <- test_card1(), card2 <- test_card2())
+  card1 <- test_card1("A title")
+  card2 <- test_card2("Another title")
+  reporter <- test_reporter(card1, card2)
+
+  reporter_1 <- Reporter$new()$append_cards(card1)
+  reporter_2 <- Reporter$new()$append_cards(card2)
+
   reporter_blocks <- reporter$get_blocks()
-  reporter_blocks2 <- append(reporter_blocks[1:3], "\\newpage")
-  reporter_blocks2 <- append(reporter_blocks2, reporter_blocks[5:8])
+  reporter_blocks2 <- append(reporter_1$get_blocks(), "\\newpage")
+  reporter_blocks2 <- append(reporter_blocks2, reporter_2$get_blocks())
   testthat::expect_equal(reporter$get_blocks(), reporter_blocks2)
 })
 


R/AddCardModule.R
```diff
@@ -183,7 +183,7 @@ add_card_button_srv <- function(id, reporter, card_fun) {
         arg_list <- c(arg_list, list(card = card))
       }
 
-      card <- try(do.call(card_fun, arg_list))
+      card <- try(do.call(card_fun(), arg_list))
```
m7pr, line 186: Only because `card_fun` is currently a `reactive(function)`
https://github.com/insightsengineering/teal.modules.general/blob/d44a257626f801bbc1756a094410921840281bd2/R/tm_a_regression.R#L1032-L1039

R/ReportDocument.R
```diff
@@ -0,0 +1,60 @@
...
+c.ReportDocument <- function(...){
+  # Regular c() drops classes and attributes, so we either overwrite the method
+  # or we do not use ReportDocument class, but list class.
+
+  # Does not work, if ReportDocument is the second element, and not the first.
+  # teal.reporter::report_document() -> x
+  # class(c(list(), x)) # list
+  # class(c(x, list())) # ReportDocument
+  # append(x, list(), after = 1) # ReportDocument
+  # append(x, list(), after = 0) # list()
```
m7pr, line 42: @gogonzo I like the design where we allow users to use `list()` object to manipulate the content of the reporter. If we add an extra class to S3 list, we can then control the flow of teal.reporter by checking report-card classes. If old class (`ReportCard` is provided, we maintain old implementation, if new `ReportDocument` is provided we use new implementation). However `ReportDocument` class and `attributes` can be easily lost if you run `c` where the first element is not `ReportDocument`.

```r
teal.reporter::report_document() -> x
class(c(list(), x)) # list
```

Maybe we can support regular list instead? And then `teal.reporter` flow controls 2 classes of report-cards: `ReportCard` and `list`.

R/AddCardModule.R
```diff
@@ -198,14 +198,28 @@ add_card_button_srv <- function(id, reporter, card_fun) {
...
+          }
         }
 
         if (!has_label_arg && length(input$label) == 1 && input$label != "") {
-          card$set_name(input$label)
+          if (inherits(card, "ReportCard")) {
+            card$set_name(input$label)
+          } else if (inherits(card, "ReportDocument")) {
+            #card <- c(card, list(name = paste0("# ", input$label)))
+            attr(card, "name") <- input$label
```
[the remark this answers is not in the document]
m7pr: Other idea. Instead of `attributes`, use names of elements of the list.
If element has a name 
- `name` - we use it as card's name and title
- `text` - we know it's a regular text (formatted in markdown)
- `comment` - we know it's a text provided as a comment
- `plot` - we know it's a plot and we need to print it to a file, and then put the path to the file in the final report
- `table` - we know it's a table that will be printed using knitr::kable or whatever we decide

This configuration can be provided in a file and can be extended or can be created manually by the user.

R/ReportDocument.R
gogonzo: Good point with `c()`, we can always introduce `add_content(<reportdocument>, ...)` method instead to protect the class. Let's not make this decision yet, there is still a topic about "inheritance from ReportDocument" which we need to discuss at some point.

R/AddCardModule.R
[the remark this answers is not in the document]
m7pr: Providing names to objects stored in a list would make the `block_to_code` implementation smoother.
We would be able to use names, to know how to recreate the object to a HTML.

More in this commit in comments.
https://github.com/insightsengineering/teal.reporter/pull/307/commits/a71c9e1feb3a2bddaf9a0860cfc1c6dedc21c62f

R/AddCardModule.R
gogonzo: I'm pretty sure that you should create new add_card_ui/srv dedicated only to the ReportDocument. Such card should only append the name of the card and ignore the comment. I think adding comment (appending content to the card) is a feature of the editor (which you'll implement later)

R/Previewer.R
```diff
@@ -157,13 +157,24 @@ reporter_previewer_srv <- function(id,
...
+            lapply(seq_along(cards), function(ic) {
+              previewer_collapse_item(ic, cards[[ic]]$get_name(), cards[[ic]]$get_content())
+            })
+          )
+          } else if (all(vapply(cards, inherits, logical(1), "ReportDocument"))) {
+            shiny::tags$div(
+              class = "panel-group accordion",
+              id = "reporter_previewer_panel",
+              lapply(seq_along(cards), function(ic) {
+                previewer_collapse_item(ic, attr(cards[[ic]], "name"), NULL) # TODO, substitute NULL with report content
```
gogonzo, line 174: I'm glad you're here already. If the Document is a list, it should be convertible to the markdown and markdown can be converted into other formats. So I think:

```r
<ReportDocument> |> to_markdown() |> commonmark::markdown_html()

```

gogonzo: I initialy thought that `to_markdown()` should be really named `format.ReportDocument` which would do:

```
lapply(<ReportDocument>, format_report_item)

format_report_item <- function(x, ...) {
  useMethod("format_report_item")
}


format_report_item.default <- function(x, ...) {
  x
}

format_report_item.data.frame <- function(x, ...) {
  kable(x)
}

format_report_item.rtables <- function(x, ...) { 
  rtables_to_html # markdown supports html
}

...
```

R/AddCardModule.R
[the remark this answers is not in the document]
m7pr: Checkout configuration provided as an option. This can be always changed by the app developer before running the app, if (s)he wants different rendering/ploting/printing options in the HTML output https://github.com/insightsengineering/teal.reporter/pull/307/commits/43a66803eb2f644ea81e5a82388a18c23b7488ac

R/Previewer.R
m7pr: This is just an example of a configuration provided as a list, that is set to R option, so that users can change those if they would like different plotting/printing options for pre-defined classes. This way they can also extend printing/plotting for their own classes.

https://github.com/insightsengineering/teal.reporter/pull/307/commits/43a66803eb2f644ea81e5a82388a18c23b7488ac

R/AddCardModule.R
```diff
@@ -215,3 +160,108 @@ add_card_button_srv <- function(id, reporter, card_fun) {
     })
   })
 }
+
+add_modal <- function(ns, comment = TRUE) {
```
m7pr, line 164: used in `add_card_button_srv` and `add_document_button_srv`

R/AddCardModule.R
```diff
@@ -215,3 +160,108 @@ add_card_button_srv <- function(id, reporter, card_fun) {
...
+        )
+      )
+    )
+  )
+}
+
+
+#' @rdname add_card_button
+#' @export
+add_document_button_srv <- function(id, reporter, r_card_fun) {
```
m7pr, line 224: Should this be moved to `teal`?
Uses `add_modal` from teal.reporter internals and `reporter/Reporter` class

R/DownloadModule.R
```diff
@@ -322,11 +322,16 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, showrcode, sessi
...
+      vapply(
+        reporter$get_blocks(),
+        function(e) inherits(e, "RcodeBlock"),
+        logical(1)
+      )
     )
-  )
+  } else {
+    FALSE
+  }
```
m7pr, line 336: used by `reporter_download_inputs` in `Previewer` on a `DownloadModule/Button`

R/ReportDocument.R
m7pr: Just added `edit_document_content` with 513b4991c87b37664a0f4bf305685fac23c27ed5

Trying to figure out how to use it with a module output. But the returned output is `teal_module`, not the `list()` returned by the `server` of the module.

```r
data <- teal_data()
data <- within(data, {
  require(nestcolor)
  CO2 <- CO2
})

app <- init(
  data = data,
  modules = modules(
    tm_a_regression(
      label = "Regression",
      response = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = "uptake",
          selected = "uptake",
          multiple = FALSE,
          fixed = TRUE
        )
      ),
      regressor = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variables:",
          choices = variable_choices(data[["CO2"]], c("conc", "Treatment")),
          selected = "conc",
          multiple = TRUE,
          fixed = FALSE
        )
      )
    ) |>
      edit_document_content(modify = c(1, 1, 2, 4, 3), append = c('Appended text'), after = 1)
  )
)
```

R/Previewer.R
```diff
@@ -161,7 +161,11 @@ reporter_previewer_srv <- function(id,
           class = "panel-group accordion",
           id = "reporter_previewer_panel",
           lapply(seq_along(cards), function(ic) {
-            previewer_collapse_item(ic, cards[[ic]]$get_name(), cards[[ic]]$get_content())
+            if (inherits(cards[[ic]], "ReportCard")) {
+              previewer_collapse_item(ic, cards[[ic]]$get_name(), cards[[ic]]$get_content())
+            } else if (inherits(cards[[ic]], "ReportDocument")) {
+              previewer_collapse_item(ic, attr(cards[[ic]], "name"), cards[[ic]])
```
m7pr, line 167: Can't uncollapse/see the content of the Card in the Report Previewer 

image


nor for the edited module, nor for the unedited

```r

devtools::load_all('../teal.reporter')
devtools::load_all('../teal')
devtools::load_all('.')

# general data example
data <- teal_data()
data <- within(data, {
  require(nestcolor)
  CO2 <- CO2
})

app <- init(
  data = data,
  modules = modules(
    tm_a_regression(
      label = "Regression",
      response = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = "uptake",
          selected = "uptake",
          multiple = FALSE,
          fixed = TRUE
        )
      ),
      regressor = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variables:",
          choices = variable_choices(data[["CO2"]], c("conc", "Treatment")),
          selected = "conc",
          multiple = TRUE,
          fixed = FALSE
        )
      )
    )
  )
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```

old way module - tm_a_pca

```r

devtools::load_all('../teal.reporter')
devtools::load_all('../teal')
devtools::load_all('.')

# general data example
data <- teal_data()
data <- within(data, {
  require(nestcolor)
  USArrests <- USArrests
})

app <- init(
  data = data,
  modules = modules(
    tm_a_pca(
      "PCA",
      dat = data_extract_spec(
        dataname = "USArrests",
        select = select_spec(
          choices = variable_choices(
            data = data[["USArrests"]], c("Murder", "Assault", "UrbanPop", "Rape")
          ),
          selected = c("Murder", "Assault"),
          multiple = TRUE
        ),
        filter = NULL
      )
    )
  )
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```

m7pr: Collapse out need to be changed to collapse show

image

image

R/DownloadModule.R
```diff
@@ -214,25 +214,25 @@ report_render_and_compress <- function(reporter, input_list, global_knitr, file
...
+  #   },
+  #   error = function(cond) {
+  #     print(cond)
+  #     shiny::showNotification(
+  #       ui = "Archive document error!",
+  #       action = "Please contact app developer",
+  #       type = "error"
+  #     )
+  #   }
+  # )
```
m7pr, line 235: Archiver part doesn't work yet, as `repoter$tojsondir/to_list` doesnt work yet

R/DownloadModule.R
```diff
@@ -151,6 +151,7 @@ download_report_button_srv <- function(id,
         input_list <- lapply(names(rmd_yaml_args), function(x) input[[x]])
         names(input_list) <- names(rmd_yaml_args)
         if (is.logical(input$showrcode)) global_knitr[["echo"]] <- input$showrcode
+        # this whole `downloadHandler` is not triggered right now during the Download Button action
```
m7pr, line 154: but `report_render_and_compress` can be tested using

```r
devtools::load_all(".")

reporter <- Reporter$new()

new_report_card <- report_document('## Plot', ggplot2::qplot(1:5, 1:5))

reporter$append_cards(list(new_report_card))

reporter$get_cards()
reporter$get_blocks()

input_list <- 
  list(
    author = "NEST", 
    title = "Report", 
    date = structure(20146, class = "Date"), 
    output = "html_document", 
    toc = FALSE
  )

global_knitr <- 
  list(
    echo = TRUE, 
    tidy.opts = list(width.cutoff = 60), 
    tidy = TRUE
  )

path <- 'render_output.zip'

report_render_and_compress(reporter, input_list, global_knitr, path)

```

R/Previewer.R
```diff
@@ -310,7 +317,7 @@ block_to_html <- function(b) {
   } else if (inherits(b, "TableBlock")) {
     b_table <- readRDS(b_content)
     shiny::tags$pre(
-      flextable::htmltools_value(b_table)
+      flextable::htmltools_valuee(b_table)
```
m7pr, line 320:
```suggestion
      flextable::htmltools_value(b_table)
```

R/Previewer.R
```diff
@@ -293,6 +297,9 @@ reporter_previewer_srv <- function(id,
 #' @noRd
 #' @keywords internal
 block_to_html <- function(b) {
+  if (!inherits(b, 'ContentBlock')) {
```
m7pr, line 300:
```suggestion
  if (inherits(b, 'character')) {
```

R/Reporter.R
```diff
@@ -48,7 +48,7 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
     #' reporter <- Reporter$new()
     #' reporter$append_cards(list(card1, card2))
     append_cards = function(cards) {
-      checkmate::assert_list(cards, "ReportCard")
+  checkmate::assert_list(cards, c("ReportCard", "ReportDocument")
```
m7pr, line 51:
```suggestion
     checkmate::assert_list(cards, c("ReportCard", "ReportDocument")
```

R/AddCardModule.R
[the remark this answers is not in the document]
m7pr: in the end it's the module developer that needs to provide proper markdown as an output of the module

R/Previewer.R
m7pr: We decided to let module developer to provide markdown as an input, so that we dont need to convert every object on our own

R/Previewer.R
```diff
@@ -41,9 +41,23 @@ reporter_previewer_ui <- function(id) {
...
+            )
+          ),
+          shiny::tabPanel(
+            "Modify",
+            shiny::tags$div(
+              id = "reporter_modifier",
+              shiny::uiOutput(ns("mcards"))
+            )
+          ),
+          selected = "Modify"
```
m7pr, line 60: `Modify` needs to be seen, so that textAreaInput is created so that `Preview` has a content to be converted to HTML...

R/Previewer.R
```diff
@@ -161,7 +175,12 @@ reporter_previewer_srv <- function(id,
           class = "panel-group accordion",
           id = "reporter_previewer_panel",
           lapply(seq_along(cards), function(ic) {
-            previewer_collapse_item(ic, cards[[ic]]$get_name(), cards[[ic]]$get_content())
+            if (inherits(cards[[ic]], "ReportCard")) {
+              previewer_collapse_item(ic, cards[[ic]]$get_name(), cards[[ic]]$get_content())
+            } else if (inherits(cards[[ic]], "ReportDocument")) {
+              # previewer_collapse_item(ic, attr(cards[[ic]], "name"), cards[[ic]])
+              previewer_collapse_item(ic, attr(cards[[ic]], "name"), input[[paste0('text_card', ic)]])
```
m7pr, line 182: If modifier can change the card content then this should just pull cards content, but should be reactive to changes provided in `text_card*`.

R/Previewer.R
```diff
@@ -175,6 +194,33 @@ reporter_previewer_srv <- function(id,
       }
     })
 
+    output$mcards <- shiny::renderUI({
+      reporter$get_reactive_add_card()
+      input$card_remove_id
+      input$card_down_id
+      input$card_up_id
+
+      cards <- reporter$get_cards()
```
m7pr, line 203: The same input as in `pcards`. Should both be wrapped around joint observe?

R/Previewer.R
```diff
@@ -293,6 +339,9 @@ reporter_previewer_srv <- function(id,
 #' @noRd
 #' @keywords internal
 block_to_html <- function(b) {
+  if (!inherits(b, 'ContentBlock')) {
+    shiny::HTML(commonmark::markdown_html(text = b, extensions = TRUE))
```
m7pr, line 343: `markdown::` package doesnt properly convert images without special metadata, so we decided to use `commonmark`

R/Previewer.R
```diff
@@ -287,195 +293,159 @@ reporter_previewer_srv <- function(id,
...
+                )
+              )
+            )
+          }, ignoreInit = TRUE, ignoreNULL = TRUE)
+        }
+      }
+    })
+
+    observe({
+      save_buttons <- grep("^save_edit_", names(input), value = TRUE)
```
m7pr, line 326: You can edit and save once, then on a second attempt something is off with the Modal

R/Renderer.R
```diff
@@ -191,7 +192,39 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
       args_nams <- unique(names(args))
       args <- lapply(args_nams, function(x) args[[x]])
       names(args) <- args_nams
-      do.call(rmarkdown::render, args)
+      rmd_path <- do.call(rmarkdown::render, args) # PATH IS RETURNED
+      clean_chunks <- function(input_rmd) {
```
gogonzo, line 196: instead of cleaning chunks, the chunks should be an original text with the class `code_chunk` and depending on the output type it should be either ignored or transformed.

R/Renderer.R
```diff
@@ -208,7 +241,7 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
     output_dir = character(0),
     report_type = NULL,
     # factory method
-    block2md = function(block) {
+    block2md = function(block, output) {
```
gogonzo, line 244: This method should look like this.
```r
block2md = function(block, output) {
  block_to_md(block)
}
```

It is possible that output is needed also, so:

```
switch(
output,
html_document = ...,
pdf_document = ...,
word_document = ....
)
```

R/ReportDocument.R
```diff
@@ -0,0 +1,200 @@
...
+
+  if(!grepl("eval=", params_str, fixed = TRUE)) {
+    if (params_str == "") {
+      params_str <- "eval=FALSE"
+    } else {
+      params_str <- paste0(params_str, ", eval=FALSE")
+    }
+  }
+  code_chunk_id <- paste0('code_chunk_', rlang::hash(code))
+  sprintf("```{r %s, %s}\n%s\n```", code_chunk_id, params_str, code)
```
gogonzo, line 130: Wrapping in ticks here causes problems later. Class should be rendered to desired output without need to parse-replace text.

```suggestion
  params <- list(...)
  structure(
    code,
    params = params,
    class = "code_chunk"
  )
```

R/Renderer.R
```diff
@@ -117,7 +118,7 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
 
       parsed_blocks <- paste(
         unlist(
-          lapply(blocks, function(b) private$block2md(b))
+          lapply(blocks, function(b) private$block2md(b, output))
```
gogonzo, line 121:
```suggestion
          lapply(blocks, function(b) block_to_output(b, output))
```

gogonzo (commented):
I suggest to make S3 methods for rendering and reusing them in the Renderer. Too many if-else based on the classes, it should be handled by S3 dispatch

R/Renderer.R
m7pr: In the end output was not needed as `to_flextable` does the whole job.
I changed block2md into an S3 generic in this commit. I needed to pass the `private` object/environment, so that old card types can render their content.

R/Renderer.R
m7pr: It's gonna be 

```r
lapply(blocks, function(b) block_to_output(b, private, output))
```

Unsure if that is that much of a win as `private` still needs to be used.

R/ReportDocument.R
m7pr: This issue is likely related to `clean_chunks` as well. https://github.com/insightsengineering/teal.reporter/pull/307#discussion_r2011678924

Currently, the process of generating a report output is a bit messy and works as follows:  

1. Retrieve objects from the reporter:  
   - If the object is plain text, it's directly inserted into the `.Rmd` file.  
   - If the object is an R object, it is saved as an `.Rds` file, and a corresponding chunk is added to load it.  
   - No code chunks are executed, except for those responsible for loading `.Rds` files.  

2. Render the `.Rmd` file:  
   - During rendering, the chunks that load `.Rds` files are evaluated, but afterward, they are removed from the final `.Rmd` file.  

This approach ensures that the final `.Rmd` file doesn't contain the code that loads `.Rds` objects while still producing a report output with the expected results—without needing to rerun the code.  

Since the code might not always be reproducible, I ensure that all chunks remain unevaluated during rendering (except for library imports, since those might be necessary for `print.object_class` functions). However, chunks that load saved `.Rds` files are evaluated. Later, I clean up the file accordingly.  

I believe we can delay the process of adding backticks (`` ` ``) until a later stage, like during the `block_to_markdown` phase. Then, in `block_to_html`, we can apply formatting based on a specific class if we create a structured class for this purpose.  

After thinking about it for a while, I’m convinced this approach makes a lot of sense.

R/Renderer.R
```diff
@@ -286,3 +313,85 @@ Renderer <- R6::R6Class( # nolint: object_name_linter.
...
+#' @method block_to_md character
+#' @keywords internal
+block_to_md.character <- function(block, private, ...) {
+  block
+}
+
+#' @method block_to_md gg
+#' @keywords internal
+block_to_md.gg <- function(block, private, ...) {
+  private$content2md(block)
```
gogonzo, line 373: `content2md` shouldn't be a part of the renderer class. This violates the encapsulation:
- new class (ReportDocument, gg, rtables, data.frame, ...) are independent so its conversion shouldn't be dependent on the Renderer class or anything else. 
- old classes (Renderer, Previewer) however should be able to call methods to render them to markdown of html.



I see your point in the comment that private$content2md needs to stay. There are numbers of solutions you can use but `private` shouldn't be passed for sure but `output_dir` only.

R/Renderer.R
gogonzo: `Renderer$new()` has only one occurrence which is in `report_render_and_compress` which is also a private function - I think you'll be able to simplify this process even more.

m7pr: Good suggestion. I removed Renderer for now, and moved it to `DownloadModule`. Once we remove `ReportCard` we will be able to remove few dispatches of `block_to_md` for old class like `HTMLBlock` or `TableBlock`.

R/ReportDocument.R
m7pr: Alrighty, now using a simple `code_chunk()` function that hold the code and parameters. It is then handled by `block_to_html` (wrapped in `tags$pre` tag) and `block_to_md` (adding ticks, and if output is powerpoint doing something more).

R/Renderer.R
m7pr: Good idea, changed to 

```r
lapply(blocks, function(b) block_to_md(b, output_dir = output_dir, report_type = report_type))
```

R/DownloadModule.R
```diff
@@ -322,11 +318,301 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, showrcode, sessi
...
+      sprintf(
+        "\\newpage\n\n---\n\n```{r, echo=FALSE}\ncode_block(\n%s)\n```\n",
+        shQuote(block_content_list, type = "cmd")
+      ),
+      collapse = "\n\n"
+    )
+  } else {
+    sprintf(
+      "```{r, %s}\n%s\n```\n",
+      paste(names(params), params, sep = "=", collapse = ", "),
```
gogonzo, line 531: Wow, i didn't know you can specify `sep` and `collapse` in the same time! It saves one call!

R/ReportCard.R
```diff
@@ -21,6 +21,7 @@ ReportCard <- R6::R6Class( # nolint: object_name_linter.
     #' card <- ReportCard$new()
     #'
     initialize = function() {
+      lifecycle::deprecate_warn("0.5.0", "ReportCard$new()", "report_document()")
```
gogonzo, line 24: Deprecate will throw a warning everywhere now. It is not a time to deprecate - we need to replace reporter in all teal.modules.* first

R/DownloadModule.R
m7pr: I'm not really sure if I like that : ) my whole life I have used 2 separate calls

R/ReportCard.R
m7pr: ah sorry, make sense, totally!

m7pr: I think one of the next steps would be to update code in tmg and tmc for all modules. I though we could also ask the team to help with that, so everybody can get acknowledged with the new approach

R/DownloadModule.R
```diff
@@ -322,11 +318,301 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, showrcode, sessi
...
+      line <- stringr::str_replace(line, "\\{r,", "{r")  # Remove leftover commas
+
+      new_lines <- c(new_lines, line)
+    }
+  }
+
+  writeLines(new_lines, input_rmd)
+}
+
+report_render_Rmd <- function(blocks, yaml_header, global_knitr = getOption("teal.reporter.global_knitr"), output_dir) {
```
gogonzo, line 411: I think `report_to_rmd` is more precise. Instead of `clean_chunks` regenerate Rmd without outputs.

```suggestion
report_to_rmd <- function(blocks, yaml_header, global_knitr = getOption("teal.reporter.global_knitr"), output_dir, include_echo = FALSE) {
```

R/DownloadModule.R
```diff
@@ -322,11 +318,301 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, showrcode, sessi
...
+  input_path <- file.path(
+    output_dir,
+    sprintf("input_%s.Rmd", gsub("[.]", "", format(Sys.time(), "%Y%m%d%H%M%OS3")))
+  )
+  cat(rmd_text, file = input_path)
+  input_path
+}
+
+#' @keywords internal
+block_to_md <- function(block, output_dir, report_type, ...) {
```
gogonzo, line 467: In your implementation block is turned into Rmd not md. 
```suggestion
block_to_rmd <- function(block, output_dir, report_type, ...) {
```

R/DownloadModule.R
```diff
@@ -322,11 +318,301 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, showrcode, sessi
...
+      "```{r, %s}\n%s\n```\n",
+      paste(names(params), params, sep = "=", collapse = ", "),
+      block
+    )
+  }
+}
+
+#' @method block_to_md PictureBlock
+#' @keywords internal
+block_to_md.PictureBlock <- function(block, output_dir, report_type, ...) {
```
gogonzo, line 539: what if someboy add a plot and want to keep it in a report?

R/DownloadModule.R
```diff
@@ -322,11 +318,301 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, showrcode, sessi
...
+  args <- lapply(args_nams, function(x) args[[x]])
+  names(args) <- args_nams
+
+  add_eval_false(input_path)
+  do.call(rmarkdown::render, args)
+  clean_chunks(input_path)
+  output_dir
+}
+
+add_eval_false <- function(input_rmd) {
```
gogonzo, line 360: instead of this you can pass `block_to_rmd.code_chunk(..., eval = FALSE)`

gogonzo (commented):
There should be no string replacement when handling code_chunk class.

R/DownloadModule.R
m7pr: good idea

R/DownloadModule.R
m7pr: >  Instead of clean_chunks regenerate Rmd without outputs.

What do you mean by that?

m7pr: >  include_echo = FALSE

Do you want to pass `echo=FALSE` as a parameter to all chunks?

R/DownloadModule.R
m7pr: We could check if objects have an attribute `attr("keep_in_rmd")` or `attr("keep_in_report")` and leave those objects as `.rds` files that area loaded from a chunk. For the plot we can substitute the chunk and loading with writing to a file and including the image with `knitr::include_img`, but that's still saving a file and having one chunk to restore it.

R/DownloadModule.R
gogonzo: > > Instead of clean_chunks regenerate Rmd without outputs.

> What do you mean by that?

in `report_render` you can `report_render_Rmd` to create a Rmd with outputs (as code_chunks) and then remove these code chunks from Rmd file.

```r
  input_path <- report_render_Rmd(blocks, yaml_header, global_knitr, output_dir)
  ...
  do.call(rmarkdown::render, args)
  clean_chunks(input_path)
```

instead you can do:

```r
  input_path <- report_render_Rmd(blocks, yaml_header, global_knitr, output_dir, include_echo = TRUE)
  ...
  do.call(rmarkdown::render, args)
  report_render_Rmd(blocks, yaml_header, global_knitr, output_dir, include_echo = FALSE)
```

m7pr: Brilliant 🔥

R/DownloadModule.R
m7pr: 9ae2e46aeaa77c72ea455e9c6767ca305cfa7e6e

R/DownloadModule.R
m7pr: You mean like this?
https://github.com/insightsengineering/teal.reporter/pull/307/commits/e6a3c6ad75055658e27a15f5b073cbbc1c0fc4ca

gogonzo: Yup

R/DownloadModule.R
m7pr: I tried that in here a2a103b474459f3511ecda382961dbab79051a8d

R/DownloadModule.R
```diff
@@ -190,10 +190,9 @@ report_render_and_compress <- function(reporter, input_list, global_knitr, file
   }
 
   yaml_header <- as_yaml_auto(input_list)
-  renderer <- Renderer$new()
 
   tryCatch(
-    renderer$render(reporter$get_blocks(), yaml_header, global_knitr),
+    output_dir <- report_render(reporter$get_blocks(), yaml_header, global_knitr),
```
gogonzo, line 195: `report_render_and_compress` is also weird. Here is my take on it:
 - `reporter_previewer_srv` calls report_render_and_compress which has shiny::showNotification and uses (shiny) `input_list`. This is a bad code separation. `report_render_and_compress` as an utility function shoul be free of shiny related stuff so ... (see next bullet point)
 - reporter_previewer_srv should check input list and showNotification
 
     ```r
        # from reporter_previewer_srv
        shiny::showNotification("Rendering and Downloading the document.")
        shinybusy::block(id = ns("download_data_prev"), text = "", type = "dots")
        input_list <- lapply(names(rmd_yaml_args), function(x) input[[x]])
        names(input_list) <- names(rmd_yaml_args)
        if (is.logical(input$showrcode)) global_knitr[["echo"]] <- input$showrcode
        report_render_and_compress(reporter, input_list, global_knitr, file)
        
        # MOVED CODE
        if (
            identical("pdf_document", input$output) &&
              inherits(try(system2("pdflatex", "--version", stdout = TRUE), silent = TRUE), "try-error")
          ) {
            shiny::showNotification(
              ui = "pdflatex is not available so the pdf_document could not be rendered. Please use other output type.",
              action = "Please contact app developer",
              type = "error"
            )
            stop("pdflatex is not available so the pdf_document could not be rendered.")
          } 
          
          
     ```
     
  - `report_render_and_compress` argument `input_list` should be renamed to `header_args` or even `yaml_header` should be already provided.
  - With couple more modifications `report_render_and_compress` could be removed and instead you could call `report_render` directly in `reporter_previewer_srv` (with error handling)

R/DownloadModule.R
```diff
@@ -322,11 +318,260 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, showrcode, sessi
...
+
+  do.call(rmarkdown::render, args)
+  file.remove(input_path)
+
+  # Create .Rmd file
+  report_to_rmd(blocks, yaml_header, global_knitr, output_dir, include_echo = FALSE) #TODO remove eval=FALSE also
+  output_dir
+}
+
+report_to_rmd <-
```
gogonzo, line 364: I suggested the name `report_to_rmd` but function isn't very similar to `block_to_rmd` as it returns a path. However if you'd return a rmd_text instead then you have a nice `to_rmd` variant - with this I mean that you can have:

```r
to_rmd <- function(x, ...) {
  useMethod("to_rmd")
}
to_rmd.default <- function()
to_rmd.character <- function()
to_rmd.code_chunk <- function()
to_rmd.Reporter <- function()
to_rmd.TextBlock <- function()
...
```

gogonzo (commented):
I've spotted more code to organise

R/DownloadModule.R
m7pr: nice idea, let me think about it

R/DownloadModule.R
m7pr: thanks, will try to play around with it

R/DownloadModule.R
m7pr: Renamed those in this commit https://github.com/insightsengineering/teal.reporter/pull/307/commits/680b1c5fcbfac59d63d64ce9d67e13d7cac99225

R/DownloadModule.R
m7pr: So far I ended up with this 6db3259e836cb192203b9c4ec658a262d9df5b2c
The errors and warnings from `report_render_and_compress` are not yet displayed on the UI.
Will try to update the code so `report_render` is used in `reporter_previewer_srv` and notifications are handled and shown

gogonzo (commented):
Edit doesn't seem to work. Nothing happens after clicking "edit" button

gogonzo (commented):
Please fix tests, start writing documentation and examples.

R/Reporter.R
```diff
@@ -48,11 +48,27 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
+        lapply(new_order, function(name) private$cards[[name]]$clone()),
+        new_order
+      )
+      invisible(self)
+    },
+    set_card_content = function(card_name, card_content) {
+      card_id <- which(names(private$cards) == card_name)
+      private$cards[[card_id]] <- card_content
+      invisible(self)
+    },
```
m7pr, line 71: Need some help here maintaining names

```r
library(ggplot2)
library(rtables)

card1 <- ReportCard$new()

card1$append_text("Header 2 text", "header2")
card1$append_text("A paragraph of default text")
card1$append_plot(
  ggplot(iris, aes(x = Petal.Length)) + geom_histogram()
)
card1$set_name('Card1')

reporter <- Reporter$new()
reporter$append_cards(list(card1))

card2 <- ReportCard$new()

card2$append_text("Header 2 text", "header2")
card2$append_text("A paragraph of default text")
lyt <- analyze(split_rows_by(basic_table(), "Day"), "Ozone", afun = mean)
table_res2 <- build_table(lyt, airquality)
card2$append_table(table_res2)
card2$set_name('Card2')

reporter$set_card_content("Card1", card2)
reporter$get_cards()[[1]]$get_name()

```

```r
> reporter$get_cards()
$Card1
<ReportCard>
  Public:
    append_content: function (content) 
    append_html: function (content) 
    append_metadata: function (key, value) 
    append_plot: function (plot, dim = NULL) 
    append_rcode: function (text, ...) 
    append_table: function (table) 
    append_text: function (text, style = TextBlock$new()$get_available_styles()[1]) 
    clone: function (deep = FALSE) 
    from_list: function (card, output_dir) 
    get_content: function () 
    get_metadata: function () 
    get_name: function () 
    initialize: function () 
    reset: function () 
    set_name: function (name) 
    to_list: function (output_dir) 
  Private:
    content: list
    deep_clone: function (name, value) 
    dispatch_block: function (block_class) 
    metadata: list
    name: Card2

> names(reporter$get_cards())
[1] "Card1"
> reporter$get_cards()[[1]]$get_name()
[1] "Card2"
```

R/DownloadModule.R
m7pr: I substituted `report_render_and_compress` with it's content in `reporter_previewer_srv` 6685163cabe4bb7793bc6d0dc82688826aaaddb9

However I don't think we can remove `report_render_and_compress` as it's used in `download_report_button_srv`.
`download_report_button_srv` is used in `simple_reporter_srv`. `simple_reporter_srv` will be deprecated, but not yet removed. I think we should bring back shiny notifications to `report_render_and_compress` so that whatever old module still uses `simple_reporter_srv` it has the same behavior as on main branch.

R/Previewer.R
```diff
@@ -279,203 +285,168 @@ reporter_previewer_srv <- function(id,
...
+            inherits(try(system2("pdflatex", "--version", stdout = TRUE), silent = TRUE), "try-error")) {
+          shiny::showNotification(
+            ui = "pdflatex is not available so the pdf_document could not be rendered. Please use other output type.",
+            action = "Please contact app developer",
+            type = "error"
+          )
+          stop("pdflatex is not available so the pdf_document could not be rendered.")
+        }
+
+        report_render_and_compress(reporter, yaml_header, global_knitr, file)
```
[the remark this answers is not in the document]
m7pr: I think it's connected to this comment https://github.com/insightsengineering/teal.reporter/pull/307#discussion_r2023720951
I will check if we can actually reuse `download_report_button_srv` in here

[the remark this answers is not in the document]
m7pr: Tried that on this branch, but so far the change does not produce the report on a download but a weird `download.htm` file https://github.com/insightsengineering/teal.reporter/pull/309/files

m7pr:
I updated the opening comment of this PR so it contains an example with 2 adjusted modules to work with ReportDocument and with one old module that works on ReportCard object https://github.com/insightsengineering/teal.reporter/pull/307#issue-2875006725

The TODO list (besides open threads in this PR):

- check bookmarking sizes with `shinyApp(..., enableBookmarking = "server")`
- provide example with a module where new `report_card` function is modified
- work on the edit button so it allows to edit the report document in reporter previewer
- documentation - rewrite documentation for `report_document()` and review `Reporter` methods documentation
- add tests for new `ReportDocument` class
- fix 2 failing tests
- maybe craft a vignette explaining how to use new `ReportDocument` class
- figure out if we can reuse duplicated from in `download_report_button_srv` in `reporter_previewer_srv` https://github.com/insightsengineering/teal.reporter/pull/307#discussion_r2022174598
- fixing names after content of the card is changed https://github.com/insightsengineering/teal.reporter/pull/307#discussion_r2023664386

m7pr:
Marked the PR ready for the review just to start CI/CD jobs. 2 tests in testthat folder are failing locally

R/Reporter.R
```diff
@@ -229,11 +289,20 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
+      cards <- self$get_cards()
+      for (i in seq_along(cards)) {
         # we want to have list names being a class names to indicate the class for $from_list
-        card_class <- class(card)[1]
+        card_class <- class(cards[[i]])[1]
         u_card <- list()
-        u_card[[card_class]] <- card$to_list(output_dir)
+        if (card_class == "ReportDocument") {
+          tmp <- tempfile(fileext = ".rds")
+          suppressWarnings(saveRDS(cards[[i]], file = tmp))
```
llrs-roche, line 299: I now realize that saveRDS is used, and it just occurred to me that some S4 objects are not easily recovered (not sure if it happens on MAE or not).

R/ReportDocument.R
```diff
@@ -0,0 +1,159 @@
...
+  input_objects <- list(...)
+  attrs <- attributes(input_objects[[1]])
+  objects <- do.call(c, lapply(input_objects, unclass))
+  attributes(objects) <- attrs
+  objects
+}
+
+#' @rdname report_document
+#' @export
+`[.ReportDocument` <- function(x, i) {
```
llrs-roche, line 69: `[` and `[[` are complex, and users are very used to them. What will happen with `report_document[1, 1]` or `report_document[1, 1, drop = TRUE]`? Would it be better to use `[[` method?

R/ReportDocument.R
```diff
@@ -0,0 +1,159 @@
...
+#'
+#' @param code A character string containing the R code to be wrapped in the chunk.
+#' @param ... Additional named parameters to be included inside `{r}`.
+#'
+#' @return A formatted character string representing an R Markdown code chunk.
+#' @examples
+#' code_chunk("x <- 1:10", echo = TRUE, message = FALSE)
+#' @export
+#' @rdname code_output
+code_chunk <- function(code, ...) {
```
llrs-roche, line 126: Maybe it is worth to add the language R by default in a helper/additional function? I suggests this because it is not included by default:

```r
code_chunk("x <- 1:10", echo = TRUE, message = FALSE)
[1] "x <- 1:10"
attr(,"params")
attr(,"params")$echo
[1] TRUE

attr(,"params")$message
[1] FALSE

attr(,"class")
[1] "code_chunk"
```

R/Previewer.R
```diff
@@ -157,12 +155,33 @@ reporter_previewer_srv <- function(id,
...
-          })
+        tags$div(
+          tags$div(
+            class = "panel-group accordion",
+            id = "reporter_previewer_panel",
+            setNames(
+              lapply(names(cards), function(card_name) {
+                if (inherits(cards[[card_name]], "ReportCard")) {
+                  previewer_collapse_item(card_name, cards[[card_name]]$get_content())
+                } else if (inherits(cards[[card_name]], "ReportDocument")) {
```
llrs-roche, line 166: There are several uses of inherits... maybe code is more readable with a `is.ReportCard` and `is.ReportDocument`?

R/DownloadModule.R
```diff
@@ -172,110 +172,44 @@ download_report_button_srv <- function(id,
...
-    },
-    error = function(cond) {
-      print(cond)
-      shiny::showNotification(
-        ui = "Copying file error!",
-        action = "Please contact app developer",
-        type = "error"
-      )
-    }
+    file.copy(temp_zip_file, file),
```
llrs-roche, line 209: Here tryCatch is used but I would also check the output of `file.copy` sometimes they fail but without warning or error (but I might confuse that with `file.move` to rename the files).

llrs-roche (commented):
This looks like a huge simplification of the code. That's great! Some feedback of running the reporter: 
I see a scroll bar on the cards that doesn't add much:
image

I had difficulties on clicking to move the cards, not sure if the area is too big or something else. 
The UI I think is nicer to have one button/card per tab, it felt natural. 

Looking into the code, I don't see issues (but I'll need some time to familiarize myself with it). I left some minor comments of suggestions/ideas as I was reading the code.

R/ReportDocument.R
m7pr: > Maybe it is worth to add the language R by default in a helper/additional function?

I dont understand. Can you elaborate?

R/Previewer.R
m7pr: could be helpful : ) indeed

m7pr:
@llrs-roche for the UI comment from the [previous comment](https://github.com/insightsengineering/teal.reporter/pull/307#pullrequestreview-2742632065), we need to summon @vedhav

R/ReportDocument.R
```diff
@@ -0,0 +1,159 @@
...
+#' To prepend, reorder, or modify a `ReportDocument`, use the `edit_report_document()` function.
+#'
+#'
+#' @examples
+#' # Create a new ReportDocument
+#' report <- report_document()
+#' class(report) # Check the class of the object
+#'
+#' # Add elements to the report
+#' report <- c(report, list("## Headline"), list("## Table"), list(summary(iris)))
```
donyunardi, line 25: Is it possible to convert the value to a list inside the function, instead of requiring the user to manually wrap it with `list()`?

For example, I want to be able to do something like this:

```r
report1 <- c(report, "## Headline", "## Table", summary(iris))
```

R/ReportDocument.R
```diff
@@ -0,0 +1,159 @@
...
+#'
+#' # Subset the report to keep only the first two elements
+#' report <- report[1:2]
+#'
+#' # Append new elements after the first element
+#' report <- append(report, c(list("## Table 2"), list(summary(mtcars))), after = 1)
```
donyunardi, line 31: Similar to my previous comment about not requiring the user to wrap the content with `list()`, it would be nice to allow something like this:

```r
report <- append(report, c("## Table 2", summary(mtcars)), after = 1)
```

Maybe we need to create `append.ReportDocument` method?

donyunardi (commented):
Nice work on this @m7pr 
I'm still playing around with this but here are a few thoughts that came up as I get myself used to this new approach.

donyunardi:
https://github.com/insightsengineering/teal.reporter/blob/204f2672fc6acb80999aa289de0f33340ad8e4c0/R/AddCardModule.R#L75-L77

Should we revisit whether the "plus" button is intuitive enough for users to understand that it adds the card to the reporter? Since there’s only one button in the module (assuming users can no longer delete a card from the module and must do it from the report previewer), should we make the button more descriptive instead of just using a "plus" sign?

R/ReportDocument.R
m7pr: @donyunardi this is because how `c(` is deisgned. `report` is a list, so to have this function work properly, all elements need to be a list. If not, they are converted to a list.


```r
> as.list(summary(iris))
[[1]]
[1] "Min.   :4.300  "
[[2]]
[1] "1st Qu.:5.100  "
[[3]]
[1] "Median :5.800  "
[[4]]
[1] "Mean   :5.843  "
[[5]]
[1] "3rd Qu.:6.400  "
```


You would not be able to do that for a regular list either. It's not caused by how `report_document` is created, but how `list`s are handled in R. And we keep `report_document` as a `list` so users are aware of it's typical behavior.

```r
l <- list("Headline")
l <- c(l, summary(iris))
> l
[[1]]
[1] "Headline"
[[2]]
[1] "Min.   :4.300  "
[[3]]
[1] "1st Qu.:5.100  "
[[4]]
[1] "Median :5.800  "
[[5]]
[1] "Mean   :5.843  "
[[6]]
[1] "3rd Qu.:6.400  "
[[7]]
[1] "Max.   :7.900  "
[[8]]
[1] "Min.   :2.000  "
```


So `c.` for `lists` runs `as.list` for every element anyway. I would not try to change this behavior to run `list` instead, which might be confusing in the later usage.

R/ReportDocument.R
m7pr: There is well known behavior and treatment for lists in R. This is just how `c` and `append` and `.[` work for lists. And this is well known to people who used lists before. Check my comment in the other thread.

https://github.com/insightsengineering/teal.reporter/pull/307#discussion_r2046384650

R/ReportDocument.R
donyunardi: As discussed, we will probably advertise `edit_report_document` function if user want to modify or append report document.

R/ReportDocument.R
donyunardi: As discussed, we will advertise `edit_report_document` function to modify or append report document.

kumamiao:
I agree the plus sign by itself down below (in the first 3 modules in the test code) may not be intuitive or apparent enough to indicate this is to add an analysis to report. But in the 4th bivariate module, this looks okay and consistent with previous design where there's a title for "Reporter". So I'm not sure the solo plus sign in the first 3 modules are simply because it's test code?
> 
> Should we revisit whether the "plus" button is intuitive enough for users to understand that it adds the card to the reporter? Since there’s only one button in the module (assuming users can no longer delete a card from the module and must do it from the report previewer), should we make the button more descriptive instead of just using a "plus" sign?

On a side note, I think I mentioned this before, but when adding cards to the reporter, users had this need to add comments/footnotes in the same time (i.e., adding some analysis details such as CI method, adding some insights, etc.), so it would be great to add it back.

m7pr:
@kumamiao 

> I agree the plus sign by itself down below (in the first 3 modules in the test code) may not be intuitive or apparent enough to indicate this is to add an analysis to report. But in the 4th bivariate module, this looks okay and consistent with previous design where there's a title for "Reporter". So I'm not sure the solo plus sign in the first 3 modules are simply because it's test code?

The 4th module is the module written in the old way. It's in the example to prove the current implemenation works for new modules and old modules. 4th bivariate module is how the reporter functionality looks right now.

m7pr:
@kumamiao 

> On a side note, I think I mentioned this before, but when adding cards to the reporter, users had this need to add comments/footnotes in the same time (i.e., adding some analysis details such as CI method, adding some insights, etc.), so it would be great to add it back.

Once the Edit card functionality is ready, I would reckon to allow to edit a card once it's added. In the Edit functionality you would be able to add a comment, and put it anywhere you want it.

R/Reporter.R
```diff
@@ -44,18 +81,59 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
-      checkmate::assert_list(cards, "ReportCard")
-      private$cards <- append(private$cards, cards)
-      private$reactive_add_card(length(private$cards))
+    #'
+    #' names(reporter$get_cards())
+    #' reporter$reorder_cards(c("Card2", "Card1"))
+    #' names(reporter$get_cards())
+    reorder_cards = function(new_order) {
+      private$cards <- setNames(
+        lapply(new_order, function(name) private$cards[[name]]$clone()),
```
m7pr, line 94: This works only for old R6 classes for now

R/ReportDocument.R
```diff
@@ -0,0 +1,159 @@
...
+#' If `NULL`, no modification is applied.
+#' @param append An object to be added to the `ReportDocument` using `append()`.
+#' The `after` parameter determines the insertion position.
+#'
+#' @examples
+#' #### edit_report_document examples ###
+#' report <- report_document(1, 2, "c")
+#'
+#' # Modify and append to the report
+#' new_report <- edit_report_document(report, modify = c(3, 1), append = "d")
```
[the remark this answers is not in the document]
gogonzo: I'm not sure if this function should even exist ;)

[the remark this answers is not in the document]
m7pr: This is a function that helps you edit/subset/modify `ReportDocument` object that guarantees the resulting object is also of class `ReportDocument`. You can always do that yourself, but you need to double check the class of the final object.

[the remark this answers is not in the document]
m7pr: Just in case fixed here https://github.com/insightsengineering/teal.reporter/pull/307/commits/2c4c0cba9dbea663eecdb00a8e2630395143d6ef

R/Reporter.R
m7pr: I think this is as desired. This will be mainly used by `ReportDocument` cards that do not have `get_name/set_name` and the name needs to remain the same, as we will use `$replace_card()` but we would like to keep the same name.

R/ReportDocument.R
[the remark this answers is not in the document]
gogonzo: report document is so simple that one can just append/replace an item with `base` functions

```r
report <- report_document(1, 2, "c")
report[[1]] <- "replacement 1"
print(report)
report <- c(
  report,
  report_document("b", "e")
)


report <- report_document(1, 2, "c")
report[[1]] <- "replacement 1"
print(report)
# [[1]]
# [1] "replacement 1"
# [[2]]
# [1] 2
# [[3]]
# [1] "c"
# attr(,"class")
# [1] "ReportDocument"

report <- c(
  report,
  report_document("b", "e")
)
print(report)
# [[1]]
# [1] "replacement 1"
# [[2]]
# [1] 2
# [[3]]
# [1] "c"
# [[4]]
# [1] "b"
# [[5]]
# [1] "e"
# attr(,"class")
# [1] "ReportDocument"
```

DESCRIPTION
```diff
@@ -37,6 +38,8 @@ Imports:
     shiny (>= 1.6.0),
     shinybusy (>= 0.3.2),
     shinyWidgets (>= 0.5.1),
+    sortable (>= 0.5.0),
```
m7pr, line 41: Currently not used at all. Cards are not sorted.

NAMESPACE
```diff
@@ -18,9 +49,11 @@ export(rmd_output_arguments)
 export(rmd_outputs)
 export(simple_reporter_srv)
 export(simple_reporter_ui)
+export(toHTML.ContentBlock)
```
m7pr, line 52: Shouldn't this be exported as `S3method`?

R/Previewer.R
```diff
@@ -30,21 +30,26 @@ NULL
...
     shiny::tagList(
-      shiny::tags$div(
-        class = "col-md-3",
-        shiny::tags$div(class = "well", shiny::uiOutput(ns("encoding")))
-      ),
-      shiny::tags$div(
-        class = "col-md-9",
+      shiny::tagList(
+        shiny::singleton(
+          shiny::tags$head(shiny::includeCSS(system.file("css/custom.css", package = "teal.reporter")))
```
m7pr, line 37: Is this a part of the removed `add_previewer_css`?

```
add_previewer_css <- function() {
  shiny::tagList(
    shiny::singleton(
      shiny::tags$head(shiny::includeCSS(system.file("css/Previewer.css", package = "teal.reporter")))
    ),
    shiny::singleton(
      shiny::tags$head(shiny::includeCSS(system.file("css/custom.css", package = "teal.reporter")))
    )
  )
}
```

R/ReportDocument.R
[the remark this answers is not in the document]
m7pr: Yeah, I think we can get rid of `edit_report_document` then. If user uses `c` and `[` and `append` while always putting new content inside `report_document()` function. This can be explained in the vignette.

R/Previewer.R
```diff
@@ -104,378 +115,232 @@ reporter_previewer_srv <- function(id,
...
         )
-      } else {
-        shiny::tags$div(
-          id = "reporter_previewer_panel_no_cards",
-          shiny::tags$p(
-            class = "text-danger mt-4",
-            shiny::tags$strong("No Cards added")
-          )
+        reporter_previewer_card_srv(
+          id = card_name,
```
averissimo, line 147: This is a bad idea to use the name verbatim here as it will create a bad DOM

image

I would argue that `Reporter$append_cards` should take the name and store it as an attribute and generate a random string as id _(using `rlang::hash` with the name)_

R/Editor.R
```diff
@@ -0,0 +1,108 @@
+editor_ui <- function(id, x) {
```
averissimo, line 1: Should we rename functions to `teal` convention of `ui_editor` instead of `editor_ui`

This is not only for this function / file but across reporter

tagging @m7pr and @gogonzo, but other feel free to chime in

R/Previewer.R
averissimo: fixed in #325

R/Reporter.R
```diff
@@ -111,83 +202,65 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
+          next # Easier to remove when ReportCard is fully deprecated
+        }
+        title <- trimws(metadata(card, "title"))
+        card_title <- if (length(title) > 0 && nzchar(title)) {
+          report_document(sprintf("# %s", title))
+        } else {
+          report_document(sprintf("# _Unnamed Card (%d)_", idx))
         }
-        blocks <- append(blocks, private$cards[[length(private$cards)]]$get_content())
+        card_with_title <- c(card_title, card)
```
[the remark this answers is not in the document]
averissimo: image

R/ReportDocument.R
```diff
@@ -0,0 +1,238 @@
...
+#' class(report)
+#'
+#' @aliases ReportDocument
+#' @name report_document
+#'
+#' @export
+report_document <- function(...) {
+  objects <- list(...)
+  structure(objects, class = c("ReportDocument"))
+}
```
averissimo, line 42: Should we use named parameters as `metadata`?

```r
report_document("## A header", "paragraph", title = "My title", author = "Spiderman")
```

averissimo:
As agreed we are closing this PR in favor of #331 that is built on top of this one.

The overhead of keeping the 2 feature branches is becoming too cumbersome with multiple function renames going on.

PR #309, opened 2025-04-01
attempt on reusing `download_report_button_srv` in `reporter_previewer_srv`

m7pr:
So far doesnt really downloads the report


m7pr:
Fixed in https://github.com/insightsengineering/teal.reporter/pull/312

PR #310, opened 2025-04-23
WIP edit module

m7pr:
Very early version of the edit module.

It is possible to 
1. edit text elements.
2. reorder elements.
3. add new text elements.

Todo:
4. once the card is saved, code elements loose their class and are not properly displayed in the previewer HTML.
5. if you delete an n-th element, all elements after this element are also deleted.
6. i think if you keep opening edit cards multiple times, at some point there is million modals and nothing longer works - like some observers should be destroyed somewhere as they keep growing.

https://github.com/user-attachments/assets/3f143adc-0837-4c24-b8ee-0cb887854ba9

### DIFF
--- a/R/EditReporterModule.R
+++ b/R/EditReporterModule.R
@@ -1,0 +1,161 @@
+report_edit_srv <- function(id, input_list_reactive, trigger_reactive, save_trigger) {
+  moduleServer(id, function(input, output, session) {
+    ns <- session$ns
+
+    current_list <- reactiveVal(NULL)
+    
+    observeEvent(trigger_reactive(), {
+      req(trigger_reactive())
+      initial_list <- input_list_reactive()
+      current_list(initial_list)
+
+      showModal(modalDialog(
+        uiOutput(ns("dynamic_list_ui")),
+        footer = tagList(
+          actionButton(ns("add_text"), "Add Text Element", icon = icon("plus")),
+          modalButton("Cancel"),
+          actionButton(ns("save"), "Save Changes", class = "btn-primary")
+        ),
+        size = "m",
+        easyClose = FALSE
+      ))
+    })
+
+    output$dynamic_list_ui <- renderUI({
+      req(current_list())
+      the_list <- current_list()
+
+      if (length(the_list) == 0) {
+        return(tags$p("The list is currently empty. Add text elements or cancel."))
+      }
+
+      list_items_ui <- lapply(seq_along(the_list), function(i) {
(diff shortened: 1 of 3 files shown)


R/EditReporterModule.R
```diff
@@ -0,0 +1,161 @@
...
+          class = "list-edit-item",
+          style = "border: 1px solid #ddd; padding: 10px; margin-bottom: 10px; background-color: #f9f9f9; display: flex; align-items: center;",
+          tags$div(style = "flex-grow: 1; margin-right: 10px;", item_content),
+          actionButton(ns(paste0("delete_item_", i)), "", icon = icon("trash"), class = "btn-danger btn-sm")
+        )
+      })
+
+      names(list_items_ui) <- paste0("item_index_", seq_along(list_items_ui))
+
+      rank_list(
```
gogonzo, line 60: `could not find function rank_list`

R/EditReporterModule.R
```diff
@@ -0,0 +1,161 @@
...
+                        value = paste(el, collapse = "\n"),
+                        rows = 2,
+                        width = "100%")
+        } else {
+          tags$div(
+            id = ns(el_id_base), 
+            style = "background-color: #eee; padding: 5px; border-radius: 3px; font-family: monospace;",
+            paste("[Non-editable object:", class(el)[1], "]")
+          )
+        }
```
gogonzo, line 48: This part should be replaced with the S3 module which should have dedicated ui_ and srv_ parts:
```suggestion
        block_to_editable_ui(...)
```

R/EditReporterModule.R
```diff
@@ -0,0 +1,161 @@
...
+
+      names(list_items_ui) <- paste0("item_index_", seq_along(list_items_ui))
+
+      rank_list(
+        labels = list_items_ui,
+        input_id = ns("rank_list_order")
+      )
+    })
+
+    observe({
```
gogonzo, line 66: No approval for `observeEvent` in the `observe

R/Previewer.R
```diff
@@ -194,6 +204,55 @@ reporter_previewer_srv <- function(id,
...
+      )
+
+    })
+
+    edited_card_content <- reactiveVal(NULL)
+    rebuild_pcards <- reactiveVal(0)
+
+    observeEvent(input$edit_selected_card, {
+      cards <- reporter$get_cards()
+      report_edit_srv(
```
gogonzo, line 240: I think it would be better to have a edit button for each card.

PR #312, opened 2025-04-30
Edit Functionality for the Report Previewer module

m7pr:
---

### Features Included in This PR

**Cards** - added support for
- **removing** (both legacy `ReportCard` and new `ReportDocument`)
- **editing** (`ReportDocument`)
- **creating new cards** (`ReportDocument`)

**Card blocks** within `ReportDocument` cards, allowed
- **editing of text blocks** 
- **deleting blocks**
- **adding new text blocks**

---

### Not Implemented (but worth considering)

- Reordering cards in the previewer.
- Reordering blocks within the `ReportDocument` edit modal.

---

### Points to Discuss

- Since Shiny doesn't support displaying multiple modals simultaneously, editing multiple text blocks can be tedious—should we explore alternatives?
- Should we allow editing of other (non-text) block types?

---

https://github.com/user-attachments/assets/480ef8de-e591-4117-8725-cdf484cf6c2e


averissimo:
Looking good @m7pr :100: 

When editing blocks, can you trigger the modal back after finishing editing, i.e. closing editing modal and re-open.

Alternative would be to renderUI the contents of the modal.

Secondary alternative would be to explore if bslib allows for this, maybe @vedhav can provide more insight.

Some UI considerations

- When adding text block it might be nice to say that it supports markdown formatting
- Consider adding a warning when deleting non-text blocks that they cannot be re-added later, only by deleting the card and adding report again
- Consider having an edit button that is disabled for the legacy cards (with a short explanation why it's disabled when hovering it)

donyunardi:
Great start, @m7pr 

Couple comments:

### Edit and Delete Buttons
* I think the the button for edit and delete should be on the right and not under the card title.
Screenshot 2025-04-30 at 3 02 50 PM

Or does it make sense to have the edit button when the accordion content is shown?
Screenshot 2025-04-30 at 3 12 54 PM

### Add New Card feature
* I am not sure I understand the purpose of the `Adding a new Card` functionality in the Report Previewer tab since user would normally start from one of the teal modules to create the report card. To me, at this point, user should only be allowed to edit the report card.

## Others:
* We should allow user to edit the card's title too.
    Screenshot 2025-04-30 at 3 00 07 PM

* Fix the code section overflow problem:

image

to:
image

R/Editor.R
```diff
@@ -0,0 +1,69 @@
...
+  )
+}
+
+#' @export
+editor_srv.ReportDocument <- function(id, x) {
+  moduleServer(id, function(input, output, session) {
+    new_content <- lapply(seq_along(x), function(i) editor_srv(i, x[[i]]))
+
+    reactive({
+      # todo: it needs to return report_document, not just list
```
m7pr, line 40:
```suggestion
```

m7pr:
@donyunardi 

- We removed `Add new Card` button.
- Fixed the width of the code chunk.
- Fixed reactivity.

What we can still consider
- moving edit/delete buttons somewhere else
- allow to edit card title

donyunardi:
> What we can still consider
> 
> moving edit/delete buttons somewhere else
> allow to edit card title

I will add these to the next increment.

kumamiao:
Thanks @m7pr for adding back the comment box when adding cards.

Agree.
> I think the the button for edit and delete should be on the right and not under the card title.

> We should allow user to edit the card's title too.

Minor comment:
Once user click on the Save button, could we close the modal by default?

Question:
In @donyunardi 's comment screenshot below, it looks like we're able to edit/delete each section for a specific card, but it does not seem to be this way when I tested. Am I pulling from the right branches? This feature would be very helpful for users to customize different sections.
`edit_onclick@redesign@main` from `teal.reporter`
`report_redesign_poc@main` from `TMG` 
`redesign@main` from `teal`

image

m7pr:
@kumamiao this branch (in teal.reporter) has been updated on Friday by @gogonzo .
The code was cleaned, but some functionality was for now removed:
- deletion of blocks in edit card modal
- adding new text block in edit card modal

We also have below in TODO https://github.com/insightsengineering/nestdevs-tasks/issues/106
- sorting blocks
- fix sorting of cards

m7pr:
There is also one known issue:

If you try to add a card with the same content that was added just before, this card will not be added because it's identical to the previous one.

PR #326, opened 2025-05-16
Improves `ReportDocument` editor

averissimo:
# Pull Request

Part of #307

### Changes description

- [x] Add new block uses `insertUI` instead of `renderUI` for new editor blocks
  - Improved performance (doesn't re-render all other blocks)
  - Keeps previous edited values
- [ ] Add title
- [ ] Add placeholder text while shiny is rendering (better UX)

[Screencast From 2025-05-16 12-40-29.webm](https://github.com/user-attachments/assets/0caf0313-8b29-4759-a423-a87a44788b96)

### DIFF
--- a/R/Editor.R
+++ b/R/Editor.R
@@ -35,11 +35,9 @@ editor_ui.ReportDocument <- function(id, x) {
 #' @export
 editor_srv.ReportDocument <- function(id, x, x_reactive) {
   shiny::moduleServer(id, function(input, output, session) {
-    output$blocks <- shiny::renderUI({
+    output$blocks <- shiny::renderUI({ # Rendered once at the beginning
       shiny::tagList(
-        lapply(names(x_reactive()), function(block_name) {
-          editor_ui(session$ns(block_name), x = x_reactive()[[block_name]])
-        })
+        lapply(names(x), function(block_name) editor_ui(session$ns(block_name), x = x[[block_name]]))
       )
     })
 
@@ -52,6 +50,13 @@ editor_srv.ReportDocument <- function(id, x, x_reactive) {
         new_blocks <- sapply(blocks_new(), function(block_name) {
           reactive_block <- shiny::reactiveVal(x_reactive()[[block_name]])
           editor_srv(block_name, x = x_reactive()[[block_name]], x_reactive = reactive_block)
+
+          insertUI(
+            sprintf("#%s", session$ns("blocks")),
+            where = "beforeEnd",
+            ui = editor_ui(session$ns(block_name), x = x_reactive()[[block_name]])
+          )
+
           shiny::observeEvent(reactive_block(), ignoreNULL = FALSE, {
             new_x <- x_reactive()
             new_x[[block_name]] <- reactive_block()


averissimo:
New feature in editor: Expandable content on reporter editor _(it doesn't make sense to show it all as there's no possible )_

[Screencast From 2025-05-16 15-59-45.webm](https://github.com/user-attachments/assets/357d30ff-c5fc-442d-8d55-cf3470614ac6)

R/ReportDocument.R
```diff
@@ -58,6 +58,71 @@ c.ReportDocument <- function(...) {
...
+  object
+}
+
+#' @export
+`id<-.ReportCard` <- function(object, ...) {
+  # Value is not needed as object can generate id
+  object$generate_id()
+  object
+}
+
```
averissimo, line 125: I'm not sure about keeping these 2 sets of functions (`id()`, `id() <-`, `label()` and `label() <-`) exported in `teal.reporter`

Pros:

- Relieve the code of some complexity as they add an API to handle the getters and setters without having `attr()` repeated call to access the attributes directly.
- Allow for better support for both `ReportDocument` and `ReportCard`

Cons:

- Not necessary outside of `teal.reporter` with 1 exception in `teal` (with a workaround possible: using `attr() <-` directly.

Alternatively, we could only register the methods and not export the generic (that is, remove the `@export` hint from the generic)

gogonzo (commented):
Let's have a call Tomorrow. I need to review this. I think we are prisoners of the design. label is a title of the document and it could be:
# <title> in markdown format
title: <title> in yaml header of .Rmd or .qmd 

If we decide that document we are building is an evaluable markdown (qmd or Rmd ) then it makes sense to have a headerattribute in the ReportDocument. title (label) should be one of the header attributes.
----
Considering `id` I think we should avoid this to be an attribute - can we completely avoid it? By avoiding I mean to set `id` when adding a card as "name" of the item in card list.

R/Editor.R
```diff
@@ -2,109 +2,204 @@ editor_ui <- function(id, x) {
...
+          footer = shiny::tagList(
+            shiny::actionButton(session$ns("edit_save"), label = "Save"),
+            shiny::modalButton("Close")
+          )
+        )
+      )
 
-    shiny::observeEvent(debounced_content(), {
-      x_reactive(debounced_content())
+      result <- srv_report_document_editor(random_editor_id, template_card)
```
gogonzo, line 148: New server instance per button click is not good. Maybe show ui n-times but call server only once?

averissimo: I'll try to find a way to better manage the data. I'm not a fan of the previous design and this is not ideal, maybe there's a middle ground!

gogonzo: `srv_report_document_editor` can be called once as it gets `template_card` (reactiveVal). Let's call `ui/srv_report_document_editor` with const id, like "editor". UI can be rerendered many times but calling server multiple times (registering observers) can cause unexpected bugs.

gogonzo (commented):
After editing previewer still show an old content

averissimo:
Functionality for editor is all there as of this comment.

I've changed the architecture with feedback from yesterday (exception is the `id` -- see below)

- Add new and old cards to reporter
- Reorder cards
- Edit card
    - Title
    - Existing blocks
    - New blocks
- Re-edit cards
- Render preview
- Render information when no reports are present


> Re: id

I think that storing it in the cards avoids passing it through or comparing full object. I'll look into this first thing tomorrow

averissimo:
_(screencast updated)_

R/Previewer.R
```diff
@@ -124,122 +124,118 @@ reporter_previewer_srv <- function(id,
...
-      current_ids <- vapply(current_cards(), attr, character(1L), which = "id")
-
-      to_add <- reporter$get_cards()[!reporter_ids %in% current_ids]
-      to_remove <- current_cards()[!current_ids %in% reporter_ids]
-      if (length(to_add)) insert_cards(to_add)
-      if (length(to_remove)) remove_cards(to_remove)
-      current_cards(reporter$get_cards())
+      all_cards <- reporter$get_cards()
+
+      reporter_ids <- vapply(all_cards, metadata, character(1L), USE.NAMES = FALSE, which = "id")
```
gogonzo, line 157: `"id"` doesn't seem to be needed.
```suggestion
      reporter_ids <- names(all_cards)
```

R/Reporter.R
```diff
@@ -44,17 +45,25 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
+        new_cards[rds] <- lapply(new_cards[rds], self$get_template())
       }
-      cards <- mapply(private$update_attributes, card = cards, label = names(cards), SIMPLIFY = FALSE)
-      private$cards <- append(private$cards, cards)
-      shiny::isolate(private$reactive_add_card(length(private$cards)))
+
+      new_cards <- lapply(new_cards, private$update_attributes)
+
+      # Set up unique id for each card
+      names(new_cards) <- vapply(new_cards, metadata, character(1L), which = "id")
```
gogonzo, line 64: Does "id" have to be set outside. Since it is random string we can set it in this method body instead.

R/Reporter.R
```diff
@@ -132,15 +141,15 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
-    replace_card = function(id, card) {
-      if (is.character(id)) {
-        id <- which(names(private$cards) == id)
-      }
-      private$cards[[id]] <- card()
-      private$reactive_add_card(length(private$cards))
+    replace_card = function(card) {
+      checkmate::assert(private$check_append(card))
+      private$cards[[metadata(card, "id")]] <- card
+      private$hashes[[metadata(card, "id")]] <- rlang::hash(list(card, Sys.time()))
```
gogonzo, line 147: Is hash here to detect if the card has been modified? If so, then maybe it is not needed:
- `ReportDocument` is an S3, when it is modified then its pointer changes
- `ReportCard` is R6 class and since it will be superseded by `ReportDocument` we don't need to support editor feature for it. 

Given above, in previewer-module we can detect if:
- card has been added or removed by detecting that names of the card list changed
- card has been modified when any card in Reporter is not identical to the `current_cards`

gogonzo (commented):
IMO "id" element in metadata can be skipped. It is not a valid document header

R/Reporter.R
averissimo: Done, and id was removed as `metadata`

R/Editor.R
```diff
@@ -1,110 +1,216 @@
...
+      "Editable markdown block"
+    ),
+    shiny::textAreaInput(ns("content"), label = NULL, value = x, width = "100%")
   )
 }
 
 #' @export
-editor_srv.ReportDocument <- function(id, x, x_reactive) {
+srv_editor_block.character <- function(id, x) {
+  shiny::moduleServer(id, function(input, output, session) session$ns("content"))
```
gogonzo, line 48: I'm not sure how to use this now. What If I'm an app developer and I would like to override some default reporter features. For example I would like to add a specific content based on the input selection.

R/ReportCard.R
```diff
@@ -288,3 +315,13 @@ ReportCard <- R6::R6Class( # nolint: object_name_linter.
...
+
+#' @export
+length.ReportCard <- function(x) {
+  length(x$get_content())
+}
+
+# #' @export
+# `names<-.ReportCard` <- function(object, value) {
+#
+# }
```
gogonzo, line 327:
```suggestion
```

R/ReportCard.R
```diff
@@ -180,6 +180,32 @@ ReportCard <- R6::R6Class( # nolint: object_name_linter.
...
+    #'
+    #' @return `self`, invisibly.
+    #' @examples
+    #' ReportCard$new()$generate_id()$get_id()
+    set_id = function(new_id) {
+      if (identical(private$id, character(0L)) || is.null(private$id)) {
+        private$id <- new_id
+      }
+      invisible(self)
+    },
```
gogonzo, line 203: Is it used?
```suggestion
```

R/ReportDocument.R
```diff
@@ -58,6 +58,56 @@ c.ReportDocument <- function(...) {
...
+  if (is.null(which)) {
+    return(result)
+  }
+  result[[which]]
+}
+
+#' @export
+metadata.ReportCard <- function(object, which = NULL) {
+  # TODO: soft deprecate
+  result <- list(title = object$get_name(), id = object$get_id())
```
gogonzo, line 80: I think `id` not needed anymore? id is now a name of the object in the Report$cards list.

```suggestion
  result <- list(title = object$get_name())
```

R/Previewer.R
```diff
@@ -124,122 +124,124 @@ reporter_previewer_srv <- function(id,
...
+      all_cards <- reporter$get_cards()
+      reporter_ids <- names(all_cards)
+      current_cards <- Filter(Negate(is.null), shiny::reactiveValuesToList(current_cards_rvs))
+      current_ids <- names(current_cards)
+
+      # Update modified card (there is no support for ReportCard implementation)
+      lapply(
+        reporter_ids[
+          !vapply(all_cards, rlang::obj_address, character(1)) %in%
+            vapply(current_cards, rlang::obj_address, character(1))
```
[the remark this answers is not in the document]
averissimo: Although it's more readable. The `%in%` is not as efficient as it is `100` times slower.

Granted, it is in the "ms", but I we should avoid it if we can.

``` r
new_el <- function(n = 10)  {
  el <- list()
  
  for (ix in seq_len(n)) {
    el[[ix]] <- list(
      paste0("name", ix),
      paste0("description", ix),
      "## Title",
      "## yada",
      iris = iris[sample(1:150, 1), ],
      # mtcars = mtcars[sample(1:32, 1), ],
      ggplot2::ggplot(data = iris, mapping = ggplot2::aes(x = Sepal.Length, y = Petal.Length)) +
        ggplot2::geom_point()
    )
  }
  el
}

main <- list(cards = lapply(1:10, new_el))
current_cards_rvs <- shiny::reactiveValues()
for (ix in sample(seq_along(main$cards), size = min(length(main), 60))) {
  current_cards_rvs[[sprintf("%d", ix)]] <- main$cards[[ix]]
}

all_cards <- main$cards
current_cards <- shiny::isolate(shiny::reactiveValuesToList(current_cards_rvs))

comp1 <- \() !current_cards %in% all_cards

comp2 <- \() !vapply(current_cards, rlang::obj_address, character(1)) %in%
  vapply(all_cards, rlang::obj_address, character(1))

bench::mark(
  identical = comp1(),
  addr = comp2(),
  min_iterations = 10,
  time_unit = "ms"
)
#> # A tibble: 2 × 6
#>   expression     min  median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>   <dbl>   <dbl>     <dbl> <bch:byt>    <dbl>
#> 1 identical  26.1    27.4         34.6     854KB      0  
#> 2 addr        0.0180  0.0209   46035.         0B     18.4
```

<sup>Created on 2025-05-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/Reporter.R
```diff
@@ -469,6 +487,16 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
+    },
+
+    # @description Check if a card can be appended to the reporter.
+    #
+    # @param card (`ReportDocument`) ReportDocument card.
+    # @param card_id (`character(1)`) card_id of the card.
+    # @return `TRUE` if card can be safely added, otherwise, `FALSE`.
+    check_append = function(card, card_id) {
+      ix <- if (length(card_id) == 0L) rep(TRUE, length(private$cards)) else names(private$cards) != card_id
+      (!metadata(card, "title") %in% vapply(private$cards[ix], metadata, character(1L), which = "title")) || return("Card with this name already exists")
```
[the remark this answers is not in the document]
averissimo: I thought unique names is a requirement, when talking with @m7pr 

I agree that this is not needed

R/Previewer.R
```diff
@@ -124,122 +124,124 @@ reporter_previewer_srv <- function(id,
...
+        "No reports have been added yet."
+      )
+    ),
+    bslib::accordion(id = ns("reporter_cards"), open = FALSE)
+  )
+}
+
+reporter_previewer_cards_srv <- function(id, reporter) {
+  moduleServer(id, function(input, output, session) {
+    current_cards_rvs <- shiny::reactiveValues()
```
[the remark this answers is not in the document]
averissimo: I did, but then I was worried about dropping the list element that is being used in the `reporter_previewer_card_srv` call.

I thought that even though the `reactiveVal` is removed from the list, the reference would still be used (and kept in memory) in the call, so we would still need to account for the same complexity.

If this was the case, better to use a `shiny` data structure that holds a list.

R/Reporter.R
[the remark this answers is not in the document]
gogonzo: I think it was related with the fact that `id` of the card was it's title and caused some namespace problems.

R/Previewer.R
```diff
@@ -124,122 +124,124 @@ reporter_previewer_srv <- function(id,
...
-      lapply(names(cards), function(card_name) {
-        card_id <- attr(cards[[card_name]], "id", exact = TRUE)
+    shiny::observeEvent(insert_queue_rv(), {
+      lapply(insert_queue_rv(), function(card_id) {
         bslib::accordion_panel_insert(
           id = "reporter_cards",
-          reporter_previewer_card_ui(id = session$ns(card_id), card_name = card_name)
+          reporter_previewer_card_ui(id = session$ns(card_id), card_id = card_id)
         )
+        current_cards_rvs[[card_id]] <- reporter$get_cards(card_id)[[1]]
```
gogonzo, line 183: Change in `$get_cards` to introduce an index is not needed. Returning a list is good enough to not introduce new lines of code in `$get_cards`. `get_cards(index)` is used here only  


```suggestion
        current_cards_rvs[[card_id]] <- reporter$get_cards()[[card_id]]
```

R/Editor.R
averissimo: I'm not sure I understand what you mean here, but we could partially revert to old behaviour to add more control on these srv functions by using `eventReactive()`

See POC below that would allow for greater flexibility on the app developer on what to return

```r
library(shiny)

ui_editor <- function(id) {
  ns <- NS(id)
  tagList(
    textInput(ns("caption"), "Caption", "Data Summary")
  )
}

srv_editor <- function(id, val_r) {
  moduleServer(id, function(input, output, session) {
    eventReactive(input$caption, {
      logger::log_info("Updating value {id}")
      sprintf("prefix... %s", input$caption)
    })
  })
}

ui_editor_main <- function(id) {
  ns <- NS(id)
  tagList(
    ui_editor(ns("yada")),
    ui_editor(ns("foo")),
    actionButton(ns("save"), "Save"),
    verbatimTextOutput(ns("value"))
  )
}

srv_editor_main <- function(id) {
  moduleServer(id, function(input, output, session) {
    
    val_r <- reactiveVal(list(yada = "", foo = ""))
    e_r <- list()
    e_r$yada <- srv_editor("yada", val_r)
    e_r$foo <- srv_editor("foo", val_r)

    observeEvent(input$save, {
      new_x <- val_r()
      for (ix_name in names(e_r)) {
        new_x[[ix_name]] <- e_r[[ix_name]]()
      }
      val_r(new_x)
    })
    
    output$value <- renderPrint({ val_r() })
  })
}

ui <- fluidPage(
  ui_editor_main("main")
)

server <- function(input, output) {
  srv_editor_main("main")
}

shinyApp(ui, server)
```

gogonzo: Yes, I think so it is better this ⬆️ way. We will `@export` a generic `ui_editor` and `srv_editor` so anybody out there will be able to override methods from `teal.reporter`. Let's discuss this Today 

... but we will have to meet collaborators at some point again to check if we address their needs.

R/Reporter.R
```diff
@@ -44,17 +45,26 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
+        new_cards[rds] <- lapply(new_cards[rds], self$get_template())
       }
-      cards <- mapply(private$update_attributes, card = cards, label = names(cards), SIMPLIFY = FALSE)
-      private$cards <- append(private$cards, cards)
-      shiny::isolate(private$reactive_add_card(length(private$cards)))
+
+      # Set up unique id for each card
+      names(new_cards) <- vapply(new_cards, function(card) {
+        sprintf("card_%s", substr(rlang::hash(list(card, Sys.time())), 1, 8))
+      }, character(1L))
```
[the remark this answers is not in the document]
averissimo: I would prefer to use hashing or even `{uuid}` as a new dependency instead of relying on `make.unique` to avoid a new card to have the same `id` as a previous one did.

I believe this solution is tolerant to that, but I would still avoid it

R/AddCardModule.R
```diff
@@ -47,9 +47,6 @@ add_card_button_ui <- function(id) {
   # Buttons with custom css and
   # js code to disable the add card button when clicked to prevent multi-clicks
   shiny::tagList(
-    shiny::singleton(
-      shiny::tags$head(shiny::includeCSS(system.file("css/custom.css", package = "teal.reporter")))
```
averissimo, line 51: @gogonzo any reason why this is repeated throughout the `teal_reporter`?

Once the previewer's tab is added the CSS should be attached to DOM

R/Reporter.R
```diff
@@ -89,20 +98,22 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
...
+      #   lapply(new_order, function(name) {
+      #     if (inherits(private$cards[[name]], "ReportDocument")) {
+      #       private$cards[[name]]
+      #     } else {
+      #       private$cards[[name]]$clone(deep = TRUE)
+      #     }
+      #   }),
+      #   new_order
+      # )
+      # browser()
```
gogonzo, line 112: gotcha! ;]

R/Reporter.R
```diff
@@ -445,30 +441,26 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
   ),
   private = list(
     id = "",
-    cards = list(),
+    cards = NULL,
+    override_order = NULL,
```
gogonzo, line 445:
```suggestion
    cards = NULL, # reactiveValues
    override_order = character(0), # to sort cards (reactiveValues are not sortable)
```

gogonzo (approved):
Nice work 👍

Issue #315, opened 2025-05-06
Soft Deprecate old methods


donyunardi:
### Feature description

We need to soft-deprecate functions that related to the old ways to create reporter.

All the blocks are currently inherits from ContentBlock, so we think we can just create the warning from the parent class.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

m7pr:
Consider Soft Deprecating `teal.reporter::ReportCard` and `teal::TealReportCard`

m7pr:
Handled during https://github.com/insightsengineering/teal.reporter/pull/407

Issue #314, opened 2025-05-06
[Documentation] Vignette for new feature


donyunardi:
### Feature description

Need to write the documentation about the new way to build reporter card.

Content should at least include:
* explain the new card object (ReportDocument), 
* tutorial
* how they can be extended in teal

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

donyunardi:
@donyunardi to review

averissimo:
Reconsider `after` function name while writing the vignette

PR #331, opened 2025-06-03
{teal} module returns a `teal_report` class that extends from `teal_data

averissimo:
# Pull Request

Fixes:

- https://github.com/insightsengineering/teal/issues/1526

Built on top of:

- https://github.com/insightsengineering/teal.reporter/pull/307
    - _(#307 will be closed once this PR is stable)_

### Companion PRs:

-


R/teal_card.R
```diff
@@ -0,0 +1,264 @@
+#' @title `teal_card`: An `S3` class for managing `teal` reports
+#'
+#' @description `r lifecycle::badge("experimental")`
+#'
+#' The `teal_card` `S3` class provides functionality to store, manage, edit, and adjust report contents.
+#' It enables users to create, manipulate, and serialize report-related data efficiently.
+#'
+#' The `teal_card()` function serves two purposes:
+#' 1. When called with a `teal_report` object, it acts as a getter and returns the card slot
+#' 2. When called with other arguments, it creates a new `teal_card` object from those arguments
```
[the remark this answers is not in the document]
pawelru: I get why this seems elegant, and it is kind of cool in isolation. But overloading teal_card() like this makes the code harder to read. You have to stop and think: is this creating something, retrieving it, or modifying it? That kind of context-switching can slow you down, makes it more difficult and more prone to user error.

Let me give you a little extreme example to show how confusing this could be:
```
c <- car() |> car("BMW") |> car(year = 2020)
e <- engine() |> engine("V8") |> engine(power = 400) |> engine(turbo = TRUE)

engine(c, e) |> engine() |> engine(status = "on")
engine(c)[[1]]@status
```

By now, engine() is doing everything — constructing, modifying, retrieving, and even accessing internals — depending on where and how it’s used. It’s dense, but not really readable.

Keeping responsibilities separate (e.g., a clear constructor, plus get_engine(), set_engine() helpers) might not look as neat, but it makes the intent obvious and the code easier to maintain.

R/teal_card.R
```diff
@@ -0,0 +1,264 @@
...
+#' @inheritParams base::append
+#'
+#' @details The `teal_card` class supports `c()` and `x[i]` methods for combining and subsetting elements.
+#' However, these methods only function correctly when the first element is a `teal_card`.
+#' To prepend, reorder, or modify a `teal_card`, use the `edit_teal_card()` function.
+#'
+#' @examples
+#' # Create a new empty card
+#' report <- teal_card()
+#' class(report) # Check the class of the object
```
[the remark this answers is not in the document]
pawelru: I saw this a few times - in this PR as well as in the PRs in other repos. Please think where it adds a value to user-facing docs. Class names would be anyway printed out if you print the whole object

averissimo:
Sortable `shinytest2` needs to be skipped until https://github.com/rstudio/sortable/issues/123 is resolved

R/teal_report-eval_code.R
```diff
@@ -0,0 +1,41 @@
+#' @importFrom teal.code eval_code
+setMethod(
+  "eval_code",
+  signature = c(object = "teal_report"),
+  function(object, code, keep_output = NULL, code_block_opts = list(), ...) {
```
averissimo, line 5: rename to `add_to_report` ?

R/teal_card.R
```diff
@@ -0,0 +1,203 @@
...
+teal_card <- function(x, ...) {
+  if (missing(x)) {
+    structure(list(), class = "teal_card")
+  } else if (inherits(x, "teal_report")) {
+    x@teal_card
+  } else {
+    objects <- list(x, ...)
+    structure(objects, class = "teal_card")
+  }
+}
```
averissimo, line 56: Should handle case where `x` is `teal_data`/`qenv` as well. Returning `NULL` or an empty `teal_card()`

R/teal_card.R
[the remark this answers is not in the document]
m7pr: Yeah, but in `R` the setter often doesn't exist as you typically use `creator<-` shortcut.
That's why you have the same name for the creator and the setter. Then it's only getter that will be different from creator, setter and slot name. I used this functionality a couple of time in many places while writing vignettes and the documentation and I'm not getting confused, even though we change the name a couple of times already.

Personally I think the example with `car` and `engine` is still neat  and elengant, but that's my personal preference. : P

R/teal_report-eval_code.R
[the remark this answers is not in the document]
m7pr: @donyunardi should we create cards to dunno, deprecate the `eval_code` and also substitute `eval_code` with `within` in teal modules?

R/teal_report-eval_code.R
```diff
@@ -0,0 +1,41 @@
...
+      return(new_object)
+    }
+
+    if (isTRUE(keep_output)) {
+      keep_output <- setdiff(
+        ls(new_object, all.names = TRUE, sorted = TRUE), ls(object, all.names = TRUE)
+      )
+    } else if (isFALSE(keep_output)) {
+      keep_output <- NULL
+    }
```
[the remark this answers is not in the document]
gogonzo: @pawelru we are cooking something better. We try to stick to the convention related to `base::.Last.value` instead. `base` always caches the result of the last evaluation. We can do the same in `eval_code`. Output will always be cached:

```r
a <- "aa"
b <- "bb"
a
.Last.value
# [1] "aa"
``` 

For qenv it will be

```
eval_code(qenv(), {
  a <- "aa"
  b <- "bb"
  a
})$.Last.value
# [1] "aa"
```

R/teal_report-eval_code.R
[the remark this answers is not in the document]
averissimo: `eval_code` is the best/clean way to define the custom `eval_code`/`within` for the new class as this 1 change modifies `within` as well.

The only downside is that it registers the parameter as accessible in the enviornment

Dispatch sequence:

1. `within` generic
1. `within.teal_report` _(does not exist, goes to second class)_
1. `within.qenv` _(does some processing and then internally calls on `eval_code` generic)_
1. `eval_code` generic
1. `eval_code.teal_report` _(runs this implementation)_

R/teal_report-eval_code.R
```diff
@@ -0,0 +1,50 @@
+#' @inherit teal.code::eval_code
+#' @param object (`teal_report`)
+#' @param keep_output (`character` or `NULL`) Names of output objects in the environment
+#' that are will be added in the card for the reporter.
```
m7pr, line 4:
```suggestion
#' that will be added in the card for the reporter.
```

R/teal_report-eval_code.R
```diff
@@ -0,0 +1,50 @@
+#' @inherit teal.code::eval_code
+#' @param object (`teal_report`)
+#' @param keep_output (`character` or `NULL`) Names of output objects in the environment
```
m7pr, line 3: @averissimo  what if outputs are not named : P ? When I want to include summary of a model like below

```r
lm.D9 <- lm(weight ~ group)
summary(lm.D9)
```

R/teal_card.R
m7pr: @averissimo I think that's gonna be good

tests/testthat/test-Reporter.R
```diff
@@ -6,206 +6,331 @@ testthat::test_that("new returns an object of type Reporter", {
...
-  testthat::expect_identical(reporter, reporter$from_reporter(reporter))
+    names_after <- names(reporter$get_cards())
+    testthat::expect_equal(names_after, c(rev(names_before), setdiff(names_after, names_before)))
+  })
 })
 
-reporter1 <- Reporter$new()
-reporter1$append_cards(list(card1, card2))
-reporter2 <- Reporter$new()
+testthat::describe("Reporter with custom template function", {
```
averissimo, line 296: Adding to improve on coverage of using Reporter with template

R/Reporter.R
```diff
@@ -17,13 +19,57 @@ Reporter <- R6::R6Class( # nolint: object_name_linter.
     #' reporter <- Reporter$new()
     #'
     initialize = function() {
```
averissimo, line 21: Should we add a parameter on constructor that allows to define `template`s?

```suggestion
    initialize = function(template = NULL) {
      checkmate::assert_function(template, null.ok = TRUE)
      self$set_template(template)
```

DESCRIPTION
```diff
@@ -25,43 +25,55 @@ Imports:
...
+    waldo (>= 0.2.0),
     withr (>= 2.0.0)
 VignetteBuilder:
     knitr,
     rmarkdown
-RdMacros:
-    lifecycle
+Remotes:
+    insightsengineering/teal.code@main,
+    insightsengineering/teal.data@main
```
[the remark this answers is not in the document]
averissimo: Nop, leftovers from previous iteration. Removed it

R/zzz.R
```diff
@@ -10,6 +10,12 @@
     options(default_global_knitr)
   }
 
+  # Manual import instead of using backports and adding 1 more dependency
+  if (getRversion() < "4.4") {
+    `%||%` <- function(x, y) if (is.null(x)) y else x
+    assign("`%||%`", `%||%`, envir = getNamespace(pkgname))
+  }
+
```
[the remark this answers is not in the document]
averissimo: Done!

R/download.R
```diff
@@ -312,11 +259,21 @@ reporter_download_inputs <- function(rmd_yaml_args, rmd_output, showrcode, sessi
 #' @noRd
 #' @keywords internal
 any_rcode_block <- function(reporter) {
+  cards <- reporter$get_cards()
+
+  # todo: make sure code_chunk is also noticed
```
averissimo, line 264:
```suggestion
```

R/previewer_card.R
```diff
@@ -0,0 +1,46 @@
...
+
+  accordion_item <- htmltools::tagAppendChildren(
+    tag = accordion_item,
+    .cssSelector = ".accordion-header",
+    ui_previewer_card_actions(ns("actions"))
+  )
+}
+
+previewer_card_srv <- function(id, card_r, card_id, reporter) {
+  # todo: card_name should be only on the server side
```
averissimo, line 29:
```suggestion
```

R/teal_card.R
```diff
@@ -0,0 +1,354 @@
...
+  if (is.null(which)) {
+    return(result)
+  }
+  result[[which]]
+}
+
+#' @rdname metadata
+#' @export
+metadata.ReportCard <- function(object, which = NULL) {
+  # TODO: soft deprecate
```
averissimo, line 198:
```suggestion
```

tests/testthat/test-teal_card.R
```diff
@@ -203,6 +203,7 @@ testthat::describe("as.teal_card", {
     testthat::skip_if_not_installed("ggplot2")
     plot <- ggplot2::ggplot(iris) +
       ggplot2::geom_point(ggplot2::aes(x = Sepal.Length, y = Sepal.Width))
+    withr::local_options("teal.reporter.disable_teal_card_conversion" = TRUE)
```
m7pr, line 206: @averissimo do you need to update tests? some of them check `recordedplot` classes : P

averissimo: What do you mean? the purpose of this tests needs it to be `ggplot2`, thus this local_options is needed, otherwise it will convert and it's no longer identical.

m7pr (approved):
Well done. Last thing would be https://github.com/insightsengineering/teal.reporter/issues/314 and https://github.com/insightsengineering/teal.reporter/issues/315 tackled in the next PR

Issue #339, opened 2025-06-16
[Feature Request]: `c.teal_card` should preserve `metadata`


m7pr:
### Feature description

Extend `c.teal_card` so it preserves `metadata()` from all `teal_cards` and takes values from the latest `teal_card`

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

averissimo:
Better example that keeps `title` metadata from `r1`, overrides `option1` and adds new `option2`

```r
r1 <- teal_card("a doc")
metadata(r1, "title") <- "a title"
metadata(r1, "option1") <- "an option"
r2 <- teal_card("another doc")
metadata(r2, "option1") <- "overriden option"
metadata(r2, "option2") <- "another option"
result <- c(r1, r2)
metadata(result)
#> $title
#> [1] "a title"
#> 
#> $option1
#> [1] "overriden option"
#> 
#> $option2
#> [1] "another option"
```

Issue #340, opened 2025-06-17
[Feature Request]: Implement `c.teal_report` method


m7pr:
### Feature description

Should work similarly to `c.teal_data` and should preserve `@teal_card` slot.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

PR #345, opened 2025-06-23
Fix `c.teal_card` to append only new element

gogonzo:
Ultimately closes https://github.com/insightsengineering/teal.reporter/issues/339 https://github.com/insightsengineering/teal.reporter/issues/340

It is not just to append content of `y` into `x`. `y` could be created from `x` and then add some NEW stuff. When merging back, it should not append stuff which is already in `x`.

<details>
<summary>  Explaination with examples </summary>

#### Joining completely new one 
```r
a <- teal_card("Text 1", "Text 2")
b <- teal_card("Text 3")
c(a, b)
# should be: "Text 1", "Text 2", "Text 3"
```

#### Joining one which appends in the end

```r
a <- teal_card("Text 1", "Text 2")
b <- c(a, "Text 3")
c(a, b)
# should be: "Text 1", "Text 2", "Text 3"
```

#### Joining one which appends in the beginning

```r
a <- teal_card("Text 1", "Text 2")
b <- append(a, "Text 3", after = 0)
c(a, b)
# should be: "Text 3", "Text 1", "Text 2"
```

#### Joining one which modifies

```r
a <- teal_card("Text 1", "Text 2")
b <- a
b[[1]] <- "Text 11"
c(a, b)
# should be: "Text 11", "Text 2"
```

#### Joining one which removes

```r
a <- teal_card("Text 1", "Text 2")
b <- a[-1]
c(a, b)
# should be: "Text 2"
```

Given all above, it it narrows down to:
- replace entire `x` with `y` if they share some names (`y` is the one which is modified so it has a precedence)
- otherwise just c(x, y)

<details>

### DIFF
--- a/R/teal_card.R
+++ b/R/teal_card.R
@@ -92,7 +92,7 @@ c.teal_card <- function(...) {
     Reduce(
       f = function(u, v) {
         v <- as.teal_card(v)
-        if (length(names(x)) && length(names(v)) && any(names(u) %in% names(v))) {
+        if (length(names(u)) && length(names(v)) && any(names(u) %in% names(v))) {
           # if there are extra names in x it means they have been removed in y
           v
         } else {
--- a/R/teal_card.R
+++ b/R/teal_card.R
@@ -97,7 +97,7 @@ c.teal_card <- function(...) {
             v
           } else {
             warning(
-              "Appended `teal_card` doesn't removed some of the elements from previous `teal_card`.\n",
+              "Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.\n",
               "Restoring original content and adding only new items to the end of the card."
             )
             modifyList(u, v)
--- a/R/teal_card.R
+++ b/R/teal_card.R
@@ -92,9 +92,16 @@ c.teal_card <- function(...) {
     Reduce(
       f = function(u, v) {
         v <- as.teal_card(v)
-        if (length(names(u)) && length(names(v)) && any(names(u) %in% names(v))) {
-          # if there are extra names in x it means they have been removed in y
-          v
+        if (length(names(u)) && length(names(v)) && any(names(u) %in% names(v))) { # when v stems from u
+          if (all(names(u) %in% names(v))) { # nothing from `u` is removed in `v`
+            v
+          } else {
+            warning(
+              "Appended `teal_card` doesn't removed some of the elements from previous `teal_card`.\n",
+              "Restoring original content and adding only new items to the end of the card."
+            )
+            modifyList(u, v)
+          }
         } else {
           attrs <- utils::modifyList(attributes(u) %||% list(), attributes(v))
           attrs$names <- union(names(u), names(v))
--- a/R/teal_card.R
+++ b/R/teal_card.R
@@ -51,7 +51,11 @@ teal_card <- function(x, ...) {
     x@teal_card
   } else {
     objects <- list(x, ...)
-    names(objects) <- sample.int(.Machine$integer.max, size = length(objects))
+    names(objects) <- vapply(
+      sample.int(.Machine$integer.max, size = length(objects)),
+      function(x) substr(rlang::hash(list(Sys.time(), x)), 1, 8),
+      character(1)
+    )
     structure(objects, class = "teal_card")
   }
 }
--- a/R/teal_card.R
+++ b/R/teal_card.R
@@ -51,6 +51,7 @@ teal_card <- function(x, ...) {
     x@teal_card
   } else {
     objects <- list(x, ...)
+    names(objects) <- sample.int(.Machine$integer.max, size = length(objects))
     structure(objects, class = "teal_card")
   }
 }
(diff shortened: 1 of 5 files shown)


averissimo:
Should we enforce this uniqueness? I think this deviates from standard `c()` functionality.

We could suggest using `modifyList` instead for this type of functionality

I can see an use case where user wants to add a separator between sections and re-uses the same object and then is surprises as the report doesn't change

```r
pkgload::load_all("teal.reporter")

sep <- teal_card("---") # horizontal linne
report <- teal_card("## Notes", "some notes before the outputs")
report <- c(report, sep, "## Plot", ggplot2::ggplot())
report <- c(report, sep, "## Table", mtcars)
report
# should show "## notes", "some notes...", "---" , "## Plot", ggplot object, "## Table", mtcars object)
```

### `c()` functionality

``` r
c(list(a = 1), list(a = 2))
#> $a
#> [1] 1
#> 
#> $a
#> [1] 2
```

<sup>Created on 2025-06-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/teal_card.R
```diff
@@ -51,6 +51,7 @@ teal_card <- function(x, ...) {
     x@teal_card
   } else {
     objects <- list(x, ...)
+    names(objects) <- sample.int(.Machine$integer.max, size = length(objects))
```
averissimo, line 54: Suggest using `Sys.time()` and `rlang::hash()` to complement this sample

``` r
pkgload::load_all("teal.reporter")
#> ℹ Loading teal.reporter

set.seed(1)
teal_card("One")
#> $`1140350788`
#> [1] "One"
#> 
#> attr(,"class")
#> [1] "teal_card"

set.seed(1)
teal_card("Two")
#> $`1140350788`
#> [1] "Two"
#> 
#> attr(,"class")
#> [1] "teal_card"
```

<sup>Created on 2025-06-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>
```suggestion
    names(objects) <- substr(
      rlang::hash(
        list(
          Sys.time(),
          sample.int(.Machine$integer.max, size = length(objects))
        )
      ),
      1,
      8
    )
```

averissimo: Updated example with the last fix

``` r
pkgload::load_all("teal.reporter")
#> ℹ Loading teal.reporter

set.seed(1)
aa <- teal_card("One")
aa
#> $`1140350788`
#> [1] "One"
#> 
#> attr(,"class")
#> [1] "teal_card"

set.seed(1)
bb <- teal_card("Two")
bb
#> $`1140350788`
#> [1] "Two"
#> 
#> attr(,"class")
#> [1] "teal_card"

c(aa, bb)
#> $`1140350788`
#> [1] "Two"
#> 
#> attr(,"class")
#> [1] "teal_card"

# Same element
c(aa, aa)
#> $`1140350788`
#> [1] "One"
#> 
#> attr(,"class")
#> [1] "teal_card"
```

<sup>Created on 2025-06-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/teal_card.R
```diff
@@ -91,10 +92,16 @@ c.teal_card <- function(...) {
     Reduce(
       f = function(u, v) {
         v <- as.teal_card(v)
-        attrs <- utils::modifyList(attributes(u) %||% list(), attributes(v))
-        result <- c(unclass(u), v)
-        attributes(result) <- attrs
-        result
+        if (length(names(u)) && length(names(v)) && any(names(u) %in% names(v))) {
+          # if there are extra names in x it means they have been removed in y
+          v
```
averissimo, line 97: I think there is a problem with this when manually creating `teal_card`, trying to reuse the same element.

It removes all elements before the "repeated" separator.

``` r
pkgload::load_all("teal.reporter")
#> ℹ Loading teal.reporter
sep <- teal_card("---") # horizontal linne

report <- teal_card("## Notes")

report <- c(report, sep)

report <- c(report, ggplot2::ggplot())

report <- c(report, sep, "## Table")

report <- c(report, head(mtcars))

report
#> $`1629355048`
#> [1] "---"
#> 
#> $`591025333`
#> [1] "## Table"
#> 
#> $`130102288`
#>                    mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> Mazda RX4         21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> Mazda RX4 Wag     21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> Datsun 710        22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> Hornet 4 Drive    21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
#> Hornet Sportabout 18.7   8  360 175 3.15 3.440 17.02  0  0    3    2
#> Valiant           18.1   6  225 105 2.76 3.460 20.22  1  0    3    1
#> 
#> attr(,"class")
#> [1] "teal_card"
```

<sup>Created on 2025-06-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

gogonzo: I consider a following alternative to:

- _if_ `all( names(x) %in% names(y) )` _then_ `modifyList(x, y)[order(y)]`

R/teal_card.R
```diff
@@ -91,10 +92,23 @@ c.teal_card <- function(...) {
...
-        attrs <- utils::modifyList(attributes(u) %||% list(), attributes(v))
-        result <- c(unclass(u), v)
-        attributes(result) <- attrs
-        result
+        if (length(names(u)) && length(names(v)) && any(names(u) %in% names(v))) { # when v stems from u
+          if (all(names(u) %in% names(v))) { # nothing from `u` is removed in `v`
+            v
+          } else {
+            warning(
+              "Appended `teal_card` doesn't removed some of the elements from previous `teal_card`.\n",
```
averissimo, line 100:
```suggestion
              "Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.\n",
```

PR #346, opened 2025-06-30
Add seamless conversion of `teal_data` to `teal_report` object

averissimo:
### Changes description

- `teal_card(<teal_data>)` will convert existing code/outputs
    - Same applies to `as.teal_report`
- `teal_card(<teal_data>) <- ` assignment will convert `<teal_data>` object in place to `teal_report`
- Adds tests that ensure `teal_card` is the same, whether built iteratively or converted from `<teal_data>`
    - With same code

### DIFF
--- a/R/teal_report-eval_code.R
+++ b/R/teal_report-eval_code.R
@@ -29,7 +29,7 @@ setMethod(
           function(x) !inherits(x, "condition"),
           lapply(
             attr(code_elem, "outputs"),
-            function(x) structure(x, class = c("chunk_output", class(x)))
+            function(x) if (isS4(x)) x else structure(x, class = c("chunk_output", class(x)))
           )
         )
         c(items, list(this_chunk), this_outs)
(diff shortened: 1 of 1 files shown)


tests/testthat/test-teal_report-class.R
```diff
@@ -0,0 +1,40 @@
...
+
+    testthat::expect_equal(unname(teal_card(as.teal_report(td))), unname(teal_card(tr)))
+  })
+
+  it("is identical when calling as.teal_report with multiple calls", {
+    td <- Reduce(f = eval_code, init = teal.data::teal_data(), x = code)
+    tr <- Reduce(f = eval_code, init = teal_report(), x = code )
+
+    testthat::expect_equal(unname(teal_card(as.teal_report(td))), unname(teal_card(tr)))
+  })
```
averissimo, line 31: @gogonzo here's the new test

PR #353, opened 2025-07-09
Use teal card internally in `teal_reportable` feature branch

averissimo:
### Changes description

- `teal_card` is now used exclusively on internals of `Reporter`
    - Allows for edit cards using "ReportCard" in modules
    - Fix problem when doing `c(teal_card("something"), list(1, 2))` as it was only saving the first element of list
    - [x] Change examples to only use `teal_card`
- Update documentation on `ReportCard`
    - Remove deprecation as we're not officially deprecating on this release
    - Reflect major usage of `teal_card` in `Reporter`


gogonzo (approved):
👍

PR #354, opened 2025-07-10
Implement conversion of `ggplot` → `recordedplot` to save space in Report and bookmarking

### LINKED ISSUE #352: [Bug]: `saveRDS` being used in reporter creates large reports & bookmark folder
### What happened?

Problems:
- Size of report and bookmarks folders is very large on the current implementation of `teal_reportable`
- Time it takes to generate the report is affected negatively

This happens due to the way `saveRDS` saves reference objects (for example: `environments`), we can indirectly observe this by comparing the space in memory and the estimated size via `{lobstr}` package

Minimal example:

```r
sample_plot <- function() ggplot2::ggplot(iris) + ggplot2::geom_histogram(ggplot2::aes(x = Sepal.Length), binwidth = .1)

path_bare <- withr::local_tempfile()
obj0 <- sample_plot()
{
  tictoc::tic("With namespace")
  saveRDS(obj0, file = path_bare)
  tictoc::toc()
}

{
  glue::glue("obj0: (obj.si) ", object.size(obj0) |> format(units = "auto")) |> print()
  glue::glue("obj0: (lobstr) ", lobstr::obj_size(obj0) |> format(units = "auto")) |> print()
  glue::glue("obj0: (disk)   ", structure(file.size(path_bare), class = "object_size") |> format(units = "auto")) |> print()
}
#> With namespace: 0.129 sec elapsed
#> obj0: (obj.si) 12.7 Kb
#> obj0: (lobstr) 4.59 MB
#> obj0: (disk)   205 Kb
```

### sessionInfo()

```R

```

### Relevant log output

```R

```

### Code of Conduct

### Contribution Guidelines

### Security Policy

averissimo:
Fixes #352 

### Changes description

- Automated conversion of data types in `teal_card`
    - Used to simplify and compress data structures that are saved in report generation and bookmark folder
    - `ggplot2` can take multiple MB in storage while `recordedplot` has a very small footprint of a few dozes KB
        - Using `evaluate::evaluate` to record plot instead of having to draw it and then 
- Framework supports future expansions to other formats
- Adds a global option that prevents conversion
    - [ ] Q: Should we document this elsewhere? or remove it?

:information_source: Screenshots with `reporter_####.zip` contents showing file sizes

image

image


llrs-roche:
I haven't checked the code so you might have handled already, but will users be able to load/use old `.rds` report files with the new reported? 

Related to the question you made:  How easy are for useRs to readRDS the new file and continue working with the plot?  
Maybe people are used to do something like this:

```r
readRDS("file_report.rds")
library("ggplot2")
plot + my_company_theme() # Or other stuff
ggsave("updated_plot.png")
```

If it is not easy then I would document the option so that people that want to keep working on the object can find how to do so.

averissimo:
> I haven't checked the code so you might have handled already, but will users be able to load/use old .rds report files with the new reported?

First, we are already breaking previous `Report` files by saving a single `rds` instead of 1 per output per report.

This functionality is here mainly to allow to upload the `zip` file and restore the Previewer state and the Bookmark.

The `rds` files are not used in the actual report nor on the `.Rmd`.

That being said, this PR doesn't remove any support for `ggplot`, so if there is a `ggplot2` (let's say by using the R option) then it is still supported.

> Related to the question you made: How easy are for useRs to readRDS the new file and continue working with the plot? Maybe people are used to do something like this:
> 
> ```r
> readRDS("file_report.rds")
> library("ggplot2")
> plot + my_company_theme() # Or other stuff
> ggsave("updated_plot.png")
> ```
> 
> If it is not easy then I would document the option so that people that want to keep working on the object can find how to do so.

Unfortunately, that is not possible with this PR. \*BUT\* it is with the reproducible code in the `Rmd` :-) if a user wants to run R code and modify the image.

llrs-roche:
Then my "vote" is to remove the option.

R/teal_card.R
```diff
@@ -309,3 +310,35 @@ code_chunk <- function(code, ..., lang = "R") {
...
+#' @keywords internal
+.convert_teal_card_input <- function(x) {
+  if (isTRUE(getOption("teal.reporter.disable_teal_card_conversion"))) {
+    return(x)
+  }
+  if (inherits(x, "chunk_output")) {
+    structure(list(.convert_teal_card_input(x[[1]])), class = c("chunk_output"))
+  } else  if (inherits(x, "ggplot")) {
+    code <- deparse1(quote(print(x)))
+    evaluate::evaluate(code)[[2]]
```
gogonzo, line 340: I'd prefer using `grDevices::recordPlot()` directly as `evaluate::evaluate` does much than just handling a graphic.

gogonzo: I think it is a little overkill to use the `evaluate::evaluate` which: 
- parses the code, 
- checks for console-outputs, 
- checks for warnings 
... only to watch for plots. 

https://github.com/r-lib/evaluate/blob/bbd9ddbbb30ab8da56a570487095c2917caacc3d/R/watchout.R#L186

This below `summarizes` the way how evaluate handles plots
```r
{
  if (requireNamespace("ragg", quietly = TRUE) && exists("agg_record", getNamespace("ragg"))) {
    ragg::agg_record()
  } else {
    pdf(file = NULL)
  }
  dev.control(displaylist = "enable")
  dev <- dev.cur()
  on.exit(dev.off(dev))
  print(<block>)
  recordPlot()
}
```

averissimo: That looks good and it works just fine, the ragg functionality that avoids the render might justify adding it to Imports.

This will make it 22 packages in imports by my count.

gogonzo: we can skip the `ragg`. I don't mind the imports note as recursive imports length is more important.

R/teal_card.R
```diff
@@ -50,7 +50,8 @@ teal_card <- function(...) {
 #' @export
 #' @keywords internal
 teal_card.default <- function(...) {
-  x <- list(...)
+  x <- sapply(list(...), .convert_teal_card_input, simplify = FALSE, USE.NAMES = TRUE)
```
gogonzo, line 53: `lapply` preserves the names by default if `x` is a named list.
```suggestion
  x <- lapply(list(...), .convert_teal_card_input)
```

gogonzo (approved):
👍 (minor comment)

R/teal_card.R
averissimo: Removed `ragg` on 0367c29

PR #360, opened 2025-07-15
Redesign and change position of reporter options

vedhav:
Companion of https://github.com/insightsengineering/teal/pull/1562

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,6 +22,7 @@ URL: https://github.com/insightsengineering/teal.reporter,
     https://insightsengineering.github.io/teal.reporter/
 BugReports: https://github.com/insightsengineering/teal.reporter/issues
 Imports:
+    bsicons,
     bslib (>= 0.8.0),
     checkmate (>= 2.1.0),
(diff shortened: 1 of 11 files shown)


gogonzo (changes requested):
Stuff from `report_previewer_ui` has been removed but `report_previewer_srv` has been left untouched. If encoding panel is not there anymore then everything related to encoding panel should be removed. There is a lot of redundant code on the server side now.
Looking at the consequences - report_previewer_srv should probably have only `id` and `reporter` arguments which requires a deprecation.

gogonzo:
> I think we can remove other params from `reporter_previewer_srv`

This is a breaking change and a violation of the deprecation process. We shouldn't just remove arguments from public functions. What are alternatives?

> And I see that the `reporter_previewer_module` is no longer needed in `teal` as well.

It is used as `teal.reporter::reporter_previewer_ui/srv` are called

vedhav:
By remove, I mean deprecate and not use them. In `teal` where they are called, we are not making use of it.

Issue #365, opened 2025-07-29
[Bug]: Can't upload report


llrs-roche:
### What happened?

On the teal.gallery the [early-dev app](https://genentech.shinyapps.io/nest_early-dev_dev/) doesn't load reports.

Here ([early_dev.zip](https://github.com/user-attachments/files/21483753/early_dev.zip)) is the report I'm having problems uploading, it has two cards from modules (Spider and one of the AE overview ones)

When I try it a pop-up redish blob appears but doesn't provide an informative error message:

<img width="1920" height="1031" alt="Image" src="https://github.com/user-attachments/assets/bad449f7-60cc-4e68-a6b6-2e386813ff69" />

### sessionInfo()

```R

```

### Relevant log output

```R

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

averissimo:
This is not a bug per se, the filename must start with `report_`.

I've tested with the same contents of your `early_dev.zip`, but renamed and with session id updated so that the app would accept.

Pending PR that will improve the message or drop the `report_` prefix strict requirement

PR #251, opened 2024-02-19
Simple - Load Reporter - Inbuild

Polkas:
closes #81 
continuation of https://github.com/insightsengineering/teal.reporter/pull/177
linked to https://github.com/insightsengineering/teal/pull/1120 Please install this teal branch when testing the code

I created a new PR from the fork as I am no longer part of the insightengineering group. 
My work is done as a collaboration of UCB company with Roche.
insightengineering developers can edit this PR.

I followed a simple design, which was evaluated positively in [the discussion](#81).

DONE:

- New modules `report_load_srv` and `report_load_ui`; similar direct update for Previewer.
- REMOVE the `Archiver` Class as we not need it for this simplified scenario.
- Improve `to_list` and `from_list` `Reporter` methods
- Add `set_id` and `get_id` `Reporter` methods. Optionally add id to a Report which will be compared when it is rebuilt from a list. To test it in the teal example app please download a report and then add a new module or dataset to the app and try to load it back. The report can be loaded back to teal app only with the same datasets and modules. The id is added to the downloaded file name if exists.
- Improve `to_list` and `from_list` `ReportCard` methods (linked with https://github.com/insightsengineering/teal/pull/1120)
- Both already existing vignette apps are updated automatically.
- `warning(cond)` everywhere to be consistent. We should send the error/warning to the R console when STH fails.
- Add `testServer` tests for report_load_srv/report_load_ui modules.
- UI tested with all 3 bootstrap versions.

Points to consider:

- The JSON format Report representation seems to be enough, so an Archiver is unnecessary. The DB solution to save/load seems overcomplex for the project.
- No update will be required to introduce it into teal modules. Simple reporter is updated automatically and can be customized with a new teal.reporter option.
- When reloading, the Report is validated by the "report_" file name prefix later by the slot name "teal Report" and optionally by ID if it is non-empty.

Example Teal App (play with bootstrap versions, simple reporter modules, and add new data/module to confirm the report can not be then reloaded):

```r
library(teal.modules.general)
# one of c("3", "4", "5")
options("teal.bs_theme" = bslib::bs_theme(version = "4"))

data <- teal_data()
data <- within(data, {
  library(nestcolor)
  ADSL <- teal.modules.general::rADSL
})
datanames <- c("ADSL")
datanames(data) <- datanames
join_keys(data) <- default_cdisc_join_keys[datanames]

app <- teal::init(
  data = data,
  modules = teal::modules(
    teal.modules.general::tm_a_regression(
      label = "Regression",
[...]

PR #374, opened 2025-08-07
Improves on error message on loading reporter 

averissimo:
# Pull Request

- Fixes https://github.com/insightsengineering/teal.reporter/issues/365

### Changes description

- Improves error message as users may have renamed the file

### Alternative:

  - Remove naming pattern enforcement _(making sure it's a zip file)_
      - @Polkas / @gogonzo do you have insights why this is so strict? (source PR: https://github.com/insightsengineering/teal.reporter/pull/251)

https://github.com/insightsengineering/teal.reporter/blob/f776a9e217141b561b1537ca076115d2147d49ea/R/LoadReporterModule.R#L97

### DIFF
--- a/R/LoadReporterModule.R
+++ b/R/LoadReporterModule.R
@@ -134,6 +134,12 @@ load_json_report <- function(reporter, zip_path, filename) {
       }
     )
   } else {
-    shiny::showNotification("Failed to load the Reporter file.", type = "error")
+    shiny::showNotification(
+      paste(
+        "Failed to load the Reporter file.",
+        "Please make sure that the filename starts with `reporter_`."
+      ), 
+      type = "error"
+    )
   }
 }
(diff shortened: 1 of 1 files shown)


vedhav (commented):
I confirm that this fixes the issue by improving the error message. But, do we need to restrict the name like this? I think checking if it is a zip file and the zip file contains a directory with `report_*` is what the logic should be. The zip file itself could be named anything.

averissimo:
I agree @vedhav, but let's hear from the developers that introduced the check, there may be a reason for it.

Polkas:
Hi Both, have a great Friday:D

The assumption was that nobody creates this zip file manually. 
A loaded file can be one truly generated by the system, a zip with a prefix report_.
Later we check the structure, like jsonfile.

If we change the assumption that sb can prepare the input (zip) manually, then the change makes sense.

PR #414, opened 2025-10-03
400 skip headers when no code

### LINKED ISSUE #400: [Bug]: Downloading report without code keeps `Module's Code` section
### What happened?

Image

Image

### sessionInfo()

```R

```

### Relevant log output

```R

```

### Code of Conduct

### Contribution Guidelines

### Security Policy

averissimo:
# Pull Request

- Fixes #400

### Companion to
- 

### Changes description

- Adds a wrapper called `pseudo_code_chunk()` that allows to hint the reporter on elements that are not `code_chunks` but should be treated as such
  - The goal is for them to be discarded when "include code" is selected
- `code_chunk()` has an extra argument that hints the reporter that this element should be always kept


gogonzo (changes requested):
IMO users will have it hard to understand this function. I don't think this function should be exported. Consequences in teal.modules.general - and others are not nice.
Is it possible to iterate and remove unnecessary in teal only with after?

Issue #418, opened 2025-10-07
[Bug]: Plots are no longer stored in `figures` directory after downloading the report


m7pr:
### What happened?

Initially discovered in here https://github.com/insightsengineering/teal.modules.general/pull/917#discussion_r2405358797

### sessionInfo()

```R

```

### Relevant log output

```R

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

llrs-roche:
To add some context. I think this was raised previously on a different issue/PR and the conclusion was to add them back as it is helpful for users to insert the png into slides or other materials (Gdocs?). 

Feel free to edit this comment to link to Nina's comment about whether to add this feature back or not.

PR #384, opened 2025-08-28
refactor: Render images to a folder when exporting the reporter

averissimo:
# Pull Request

- Fixes #364

#### Changes description

- On report download save images to folder
    - Uses cached HTML base64 images to save time _(avoids re-rendering)_

PR #420, opened 2025-10-08
418 bring back figures storage again

m7pr:
Fixes

- https://github.com/insightsengineering/teal.reporter/issues/418

Initially introduced during

- https://github.com/insightsengineering/teal.reporter/pull/384

The issue was the `cached_html` structure was wrapped inside a `shiny::tagList` that changed the structure of this object and it was no longer possible to extract cached plots to be saved in png files. It happened during https://github.com/insightsengineering/teal.reporter/pull/405
in here

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: teal.reporter
 Title: Reporting Tools for 'shiny' Modules
-Version: 0.5.0.9009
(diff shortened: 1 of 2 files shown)


gogonzo:
What's the reason behind introducing `cached_html` argument (and slot in `Reporter`)?

PR #427, opened 2025-10-16
Reuse module's label as default Card's name

m7pr:
Companion to 

- https://github.com/insightsengineering/teal/pull/1636

### DIFF
--- a/tests/testthat/test-addCardModule.R
+++ b/tests/testthat/test-addCardModule.R
@@ -110,25 +110,34 @@ testthat::describe("add_card_button_srv", {
   })
 
   it("uses card_title as a default value for card title input", {
-    card_fun <- function(card = teal_card(), comment = NULL) {
-      card <- c(card, "## Header 2 text", "A paragraph of default text")
-      if (!is.null(comment)) {
-        card <- c(card, "### Comment", comment)
-      }
-      card
+    card_fun <- function(card = teal_card()) {
+      c(card, "## Header 2 text", "A paragraph of default text")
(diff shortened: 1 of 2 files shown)


tests/testthat/test-addCardModule.R
```diff
@@ -109,6 +109,37 @@ testthat::describe("add_card_button_srv", {
...
+      add_card_button_srv,
+      args = list(reporter = Reporter$new(), card_fun = card_fun, default_label = "My Module"),
+      expr = {
+        # Test adding the card with the default label
+        session$setInputs(`add_report_card_button` = 0)
+        session$setInputs(comment = "Test comment")
+        session$setInputs(`add_card_ok` = 0)
+
+        # Verify the card was added and has the correct title
+        testthat::expect_length(reporter$get_blocks(), 2) # 1 for title + 1 for content
```
llrs-roche: I count the `## Header 2 text`, the `paragraph of default text`, the `### comment` and the `comment`. My expectation is that we should have more than two blocks.

gogonzo (commented):
- There are three terms describing this field (label, name, title). IMO label is the least accurate. What is the best term to describe a "title/name of the raport card". `label` describes well what it is for teal_module but not for report. 
- I think argument should be named `card_title/name` instead of label and should be documented. `label` is already in `add_card_button_ui` and it refers to the button label.


Skärmavbild 2025-10-17 kl  04 09 48

gogonzo (approved):
👍

Issue #396, opened 2025-09-08
[Documentation]: Update README GIF in teal.reporter to reflect the new design


donyunardi:
### Feature description

Update README GIF in teal.reporter to reflect the new design

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

PR #428, opened 2025-10-16
showcase gif updated

m7pr:
Closes 

- https://github.com/insightsengineering/teal.reporter/issues/396


llrs-roche (commented):
The gif is good and shows the same features as the old one but with the new UI/modals.

But I think it should be updated to show how to collapse code blocks: the new button is different from were it was. 

I think it would also help to show that now the "Show R code" is on the download button modal instead of when adding a new card.

PR #431, opened 2025-10-21
Move `cached_html` argument to internal functions

averissimo:
# Pull Request

Linked to:

- https://github.com/insightsengineering/nestdevs-tasks/issues/112

### Changes description

- `srv_editor_block` and `ui_editor_block` take `...` argument to allow for method overriding.
- `cached_html` is no longer a public argument to those functions
  - That is: it becomes optional to use the cached HTML rendered when the card is being added

### Considerations

- Should we add an option to disable the caching of the reporter content when adding a card?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: teal.reporter
 Title: Reporting Tools for 'shiny' Modules
-Version: 0.5.0.9020
-Date: 2025-10-20
+Version: 0.5.0.9021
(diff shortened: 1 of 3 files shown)


averissimo:
> ui_editor_card still contains cached_html.

Does it? I think only the internal default method does. Am I seeing it wrong?

@gogonzo how do you feel of keeping it in the `ui_editor_block` / `srv_editor_block` man page?

image

PR #437, opened 2025-10-24
436 Allow for larger file upload to upload large reports

### LINKED ISSUE #436: [Feature Request]: Modify the way reports are uploaded so teal.reporter can handle reports larger than 5MB
### Feature description

The default shiny::fileInput allowed size is 5MB.
If the report generated is larger than 5MB, it won't allow to be uploaded.
One quick solution would be to modify the max.size value of the fileInput [as described here](https://mastering-shiny.org/action-transfer.html#server).
That being said, a better solution would be to not require to upload the full .zip file. That, tough, requires a different design of `LoadReporterModuleR`.

### Code of Conduct

### Contribution Guidelines

### Security Policy

osenan:
# Pull Request

Fixes #436 

Added a package option that will control the size of the `fileInput`, which [by default is 5 MB](https://mastering-shiny.org/action-transfer.html#server).
The option is a new `teal.reporter` option and by default is 10MB. With that modification we allow:

1. Control for developers to decide if they allow big reports to be uploaded.
2. Unblock the possibility of having reports larger than 5MB.

## Questions

* This implementation will modify all allowed max sizes of shiny fileInputs. Is this a problem?
* Do I need to document this option? I do not see specific documentation of other `teal.reporter` package options in vignettes or other parts.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: teal.reporter
 Title: Reporting Tools for 'shiny' Modules
-Version: 0.5.0.9022
-Date: 2025-10-24
+Version: 0.5.0.9023
+Date: 2025-10-27
 Authors@R: c(
     person("Dawid", "Kaledkowski", , "dawid.kaledkowski@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0001-9533-457X")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# teal.reporter 0.5.0.9022
+# teal.reporter 0.5.0.9023
 
 ### New features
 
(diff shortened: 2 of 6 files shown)


R/zzz.R
```diff
@@ -18,7 +18,8 @@
       author = "NEST", title = "Report",
       date = as.character(Sys.Date()), output = "html_document",
       toc = FALSE
-    )
+    ),
+    teal.reporter.max_request_size = 10 * 1024^2
```
m7pr: @osenan can you make a PR to teal, with a git branch of the same name as the branch name of this feature and update the vignette that describes options?
https://github.com/insightsengineering/teal/blob/main/vignettes/teal-options.Rmd

m7pr (approved):

PR #454, opened 2026-01-09
Fixes `teal_card<-` assignment with qenv.error to do nothing instead of throwing error

averissimo:
# Pull Request

- Part of https://github.com/insightsengineering/teal.modules.clinical/issues/1440
- Companion to https://github.com/insightsengineering/teal.modules.clinical/pull/1441 that fixes bug

### Changes description

- Add `teal_card` assignment support
- Test new funcitionality

### DIFF
--- a/R/teal_card.R
+++ b/R/teal_card.R
@@ -101,6 +101,9 @@ teal_card.qenv <- function(...) {
 #' @param value (`teal_card`) object to set in the `teal_report`.
 #' @export
 `teal_card<-` <- function(x, value) {
+  if (inherits(x, "qenv.error")) { # qenv.error does nothing
+    return (x)
+  }
   x <- methods::as(x, "teal_report")
   checkmate::assert_class(x, "teal_report")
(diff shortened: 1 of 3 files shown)


R/teal_card.R
```diff
@@ -101,6 +101,9 @@ teal_card.qenv <- function(...) {
 #' @param value (`teal_card`) object to set in the `teal_report`.
 #' @export
 `teal_card<-` <- function(x, value) {
+  if (inherits(x, "qenv.error")) { # qenv.error does nothing
+    return(x)
```
llrs-roche, line 105: Current behavior ends up in an error:
```
Error in methods::as(x, "teal_report") : 
  no method or default for coercing "qenv.error" to "teal_report"
```

Should it raise a warning? This would make it easier to detect those cases where this happens while not halting the execution.

llrs-roche (approved):
I haven't checked this helps on bringing back the tests but I think this is a good addition (assignments should always return an object).

PR #455, opened 2026-01-21
Suppress `c.teal_card` warning when joining `teal_report` objects

averissimo:
# Pull Request

Fixes warning popping up on modules using `c.teal_card` indirectly

```r
pkgload::load_all("/home/averissimo/work/roche🟦/packages📦/teal.reporter")

q1 <- teal.reporter::teal_report() |> within(1+1)
q2 <- within(q1, 2+2)
q3 <- within(q1, 3+3)
c(q2, q3)
#> Warning message:
#> Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.
#>  
#> ✅︎ code verified
#> <environment: 0x5b5d075c6b10> 🔒 
#> Parent: <environment: package:teal.reporter> 
```

### Changes Description

- Adds class to warning condition
  - Allows developers to suppress it explicitly, without hiding other warnings
- Warning should not be presented when merging `teal_report` objects
  - Makes it look as there is something wrong with the operation and on teal apps (such as `tm_g_distribution`)

### DIFF
--- a/R/teal_card.R
+++ b/R/teal_card.R
@@ -162,8 +162,11 @@ c.teal_card <- function(...) {
             v
           } else {
             warning(
-              "Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.\n",
-              "Restoring original content and adding only new items to the end of the card."
+              warningCondition(
+                "Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.\n",
+                "Restoring original content and adding only new items to the end of the card.",
+                class = "teal_card_append"
+              )
             )
             utils::modifyList(u, v)
           }
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # teal.reporter 0.6.0.9001
 
+### Miscellaneous
+
+* Prevents warning message when joining two `teal_card` objects that have items with the same id (#455).
+
 # teal.reporter 0.6.0
 
 ### New features
--- a/R/teal_report-c.R
+++ b/R/teal_report-c.R
@@ -10,7 +10,7 @@ c.teal_report <- function(...) {
   result <- NextMethod()
   l <- Filter(function(x) inherits(x, "teal_report"), list(...))
   if (length(l) > 1) {
-    teal_card(result) <- suppressWarnings(do.call(c, lapply(l, teal_card)), class = "teal_card_append")
+    teal_card(result) <- suppressWarnings(do.call(c, lapply(l, teal_card)), classes = "teal_card_append")
   }
   result
 }
--- a/R/teal_report-c.R
+++ b/R/teal_report-c.R
@@ -10,7 +10,7 @@ c.teal_report <- function(...) {
   result <- NextMethod()
   l <- Filter(function(x) inherits(x, "teal_report"), list(...))
   if (length(l) > 1) {
-    teal_card(result) <- do.call(c, lapply(l, teal_card))
+    teal_card(result) <- suppressWarnings(do.call(c, lapply(l, teal_card)), class = "teal_card_append")
   }
   result
 }


R/teal_card.R
```diff
@@ -162,8 +162,11 @@ c.teal_card <- function(...) {
             v
           } else {
             warning(
-              "Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.\n",
-              "Restoring original content and adding only new items to the end of the card."
+              warningCondition(
+                "Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.\n",
```
llrs-roche, line 166: Based on my experiments the condition needs to be triggered:

```suggestion
              w <- warningCondition(
                "Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.\n", ...)
                warning(w)
```

llrs-roche (commented):
I like having classed warnings and errors. R itself is starting to use them, and it would be very helpful if like python there are some standard error and warning classes.

I think this is related to the other PR I also reviewed about errors on qenv/teal_data/teal_report. What I'm not sure is why we wouldn't want to let the developer know that one (or some of the) card(s) added are failing.  Just thinking from the user prespective doesn't seem to help here.

I get that we don't want to add to the report something that is failing but not even knowing that is failing might be a bad decision. I would let the app-developer or module-developer know that something isn't fine.

R/teal_card.R
averissimo: It's being triggered on L164 (just before `warningCondition`), I just replaced the string with the condition

llrs-roche (approved):

averissimo:
@llrs-roche I agree, this PR intended to avoid false-positives, when merging 2 `teal_report` objects, where the `teal_card` being merged should be trusted as correct (first scenario below).

Q: Maybe we should let the warning continue to show in all cases and refine a new issue on how to best present this warning.

Context: warning triggered when first `teal_card` of join has ids that following ones don't (assumes they were deleted, but can be added as well)

### Scenario of a valid join that should not give an error

```r
q <- within(teal.reporter::teal_report(), 1 + 1)
q2 <- within(q, 2 + 2)
q3 <- within(q, 3 + 3)
c(q2, q3) # generates warning 
#> Warning message:
#> Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.
#>  
#> ✅︎ code verified
#> <environment: 0x5a5ad878f470> 🔒 
#> Parent: <environment: package:teal.reporter> 
```

Why?

Because not all names in `teal_card(q2)` are in `teal_card(q3)`.

### Scenario of a valid error

```r
q <- within(teal.reporter::teal_report(), 1 + 1)
teal_card(q) <- c(teal_card(q), "yada")
q2 <- q
teal_card(q2) <- c(teal_card(q2)[1])
c(q, q2) # generates a warning
#> Warning message:
#> Appended `teal_card` doesn't remove some of the elements from previous `teal_card`.
#>  
#> ✅︎ code verified
#> <environment: 0x5a5ad6835c28> 🔒 
#> Parent: <environment: package:teal.reporter> 


```

averissimo:
> Q: Maybe we should let the warning continue to show in all cases and refine a new issue on how to best present this warning.

@llrs-roche should we?

1. Keep PR as is 
    - Ignore when merging `teal_reporter` objects
    - Only shows when joining `teal_card` explicitly
3. Remove the suppression and merge PR

llrs-roche:
Thinking a bit more on this. Could we distinguish between these two examples when we trigger the code? I think this would be the best approach for users. Or this is too costly and complicated for us?
But I think this depends on how we see teal_report going forward, and maybe @gogonzo have some opinion on this.

averissimo:
I don't think we can right now, because we would need to keep track of the last common ancestor and that adds some overhead (if not impossible)

Maybe this warning doesn't need to exist at all.

averissimo:
Discussion continues https://github.com/insightsengineering/teal.reporter/pull/455

